PR #119, opened 2020-06-10
Gracefully shut down server with manners pkg

weihesdlegend:
## Description
improve readability and stability of gracefully shut down server with manners package

## Root Cause
good industry practice

## Covered E2E tests
manually tested

## Final Checks
- [x] You are using approved terminology

### DIFF
--- a/main/main.go
+++ b/main/main.go
@@ -1,11 +1,10 @@
 package main
 
 import (
-	"context"
+	"github.com/braintree/manners"
 	"github.com/kelseyhightower/envconfig"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


main/main.go
```diff
@@ -1,11 +1,10 @@
 package main
 
 import (
-	"context"
+	"github.com/braintree/manners"
```
sananand007, line 4: https://github.com/braintree/manners#known-issues, I guess we not need to worry about this issue rite at least for now?

weihesdlegend: right, I think we are fine now. Manners is known for its support of graceful shut down.

PR #120, opened 2020-06-14
Add google maps url to planning results

weihesdlegend:
## Description
Add google maps  url to planning results. This should be a nice enhancement for users to explore the location.

## Solution description
- Define new fields to store photo and URL meta data
- Persist in database
- Serve in upper layer solutions

## Covered E2E tests
Manually tested


solution/generate_slot_solution.go
```diff
@@ -80,7 +80,7 @@ func GenerateSlotSolution(timeMatcher *matching.TimeMatcher, location string, ev
 	}
 
 	cityCountry := strings.Split(location, ",")
-	evTags := make([]string, len(stayTimes))
```
weihesdlegend, line 83: previously this size allocation is confusing, even though at this point `len(stayTimes)==len(evTags`

PR #137, opened 2020-09-12
Create go.yml

weihesdlegend:
## Description
Add basic build and test script to enable action

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,37 @@
+name: Go
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.14
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -0,0 +1,38 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
```
sananand007, line 13: Does this have a stable version of Ubuntu, I think better to go with stable rather than bleeding edge ?

.github/workflows/go.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.14
```
sananand007, line 19: is there a way to select multiple go versions here ?

sananand007 (commented):
nit: comments

.github/workflows/go.yml
```diff
@@ -0,0 +1,38 @@
...
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
+            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+            dep ensure
+        fi
+
+    - name: Build
+      run:
+        cd main/: go build -v .
```
sananand007, line 35: can we change this to 
```
go main/build -v .
```

sananand007 (changes requested):

.github/workflows/go.yml
weihesdlegend: I think here it requires Go version >= 1.14

PR #140, opened 2020-10-01
Fix Redis place update issue

weihesdlegend:
## Description
For the places without URL after update places with new detailed search, Redis is not updating. 

## Root Cause
Production server makes way too many place details search calls

## Solution

- Make `updatePlacesDetails` method returns a list of places and update Redis. 
- Fixed issues in the Redis update method.

## Testing
Integration test procedure
- Obtain places without URL by removing URL field in the field mask
- Remove slot solutions to force the call flow to the lower-level place search
- Since search timer is not up, it forces the call flow to call the `updatePlaceDetails` function
- Check if the same place get updated URL

## Final Checks

### DIFF
--- a/graph/update_places.go
+++ b/graph/update_places.go
@@ -19,7 +19,7 @@ func detailsResultMap() *PlaceDetailsResultMap {
 
 func placeNeedUpdate(place *POI.Place) bool {
 	if len(strings.TrimSpace(place.GetURL())) == 0 || place.GetURL() == iowrappers.GoogleSearchHomePageURL {
-		logrus.Debugf("place with ID: %s need URL update", place.GetID())
+		logrus.Debugf("[REDIS URL UPDATE] place with ID: %s needs URL update", place.GetID())
 		return true
 	}
 	return false
@@ -64,5 +64,5 @@ func UpdatePlacesDetails(searcher iowrappers.SearchClient, places []POI.Place) (
 
 func updatePlaceDetails(place *POI.Place, details POI.Place) {
 	place.SetURL(details.GetURL())
-	logrus.Debugf("place with ID: %s URL updated to %s", place.GetID(), place.GetURL())
+	logrus.Debugf("[REDIS URL UPDATE] the URL of place with ID: %s has been updated to %s", place.GetID(), place.GetURL())
 }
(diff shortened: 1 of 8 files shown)


graph/update_places.go
```diff
@@ -50,9 +57,12 @@ func updatePlacesDetails(searcher *iowrappers.PoiSearcher, places []POI.Place) {
...
+		placesNeedUpdate = append(placesNeedUpdate, places[idx])
 	}
+	return placesNeedUpdate
 }
 
-func updatePlaceDetails(place *POI.Place, details maps.PlaceDetailsResult) {
-	place.URL = details.URL
+func updatePlaceDetails(place *POI.Place, details POI.Place) {
+	place.SetURL(details.GetURL())
+	logrus.Debugf("place with ID: %s URL updated to %s", place.GetID(), place.GetURL())
```
xtpitt, line 67: Add a special label in the log for easy search [TEST REDIS URL UPDATE]

graph/update_places.go
```diff
@@ -1,28 +1,33 @@
...
-func newPlaceDetailsResultMap() *PlaceDetailsResultMap {
-	return &PlaceDetailsResultMap{results: map[string]maps.PlaceDetailsResult{}}
+func detailsResultMap() *PlaceDetailsResultMap {
+	return &PlaceDetailsResultMap{results: map[string]POI.Place{}}
 }
 
 func placeNeedUpdate(place *POI.Place) bool {
-	return len(strings.TrimSpace(place.GetURL())) == 0 || place.GetURL() == iowrappers.GoogleSearchHomePageURL
+	if len(strings.TrimSpace(place.GetURL())) == 0 || place.GetURL() == iowrappers.GoogleSearchHomePageURL {
+		logrus.Debugf("place with ID: %s need URL update", place.GetID())
```
xtpitt, line 22: Add a special label in the log for easy search [TEST REDIS URL UPDATE]

xtpitt (approved):

graph/update_places.go
weihesdlegend: Done

graph/update_places.go
weihesdlegend: Done

PR #142, opened 2020-10-04
Upgrade redis client v8

weihesdlegend:
## Description

- Upgrade redis client library from v7 to v8
- Upgrade Go version to `1.15`

## Root Cause
The go-redis library version we use is outdated

## Solution
fix API changes, mostly v8 requires context to make Redis calls

## Testing
Both

## Final Checks


iowrappers/redis_client.go
```diff
@@ -26,6 +27,12 @@ const (
 	NumVisitorsPrefix      = "visitor_count"
 )
 
+var RedisClientContext context.Context
+
+func init() {
+	RedisClientContext = context.Background()
```
sananand007, line 33: why use a `Background` and not any other type of contexts ?

sananand007 (approved):
lgtm. nit: wanted to understand why are we using context.background ?

iowrappers/redis_client.go
```diff
@@ -4,10 +4,11 @@ package iowrappers
 // serving the caching needs of the Vacation Planner
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/go-redis/redis/v7"
+	"github.com/go-redis/redis/v8"
```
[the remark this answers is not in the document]
weihesdlegend: Good question, they didn't provide a feature list though. But I think from API we can see they are making some changes like `context`

iowrappers/redis_client.go
weihesdlegend: because at the moment we are not doing anything except satisfying the function signature requirement

PR #145, opened 2020-10-12
Simplify plan layout css

weihesdlegend:
## Description
One-day plan page layout was broken on mobile. Root cause is column width is not adjusting per device type.

## Solution
Use Boostrap's auto-column layout feature we can specify how many columns we want for each row. Then let Bootstrap's magic fix the breakpoints per device type.

## Testing
Tested on Heroku preview app and mobile.

## Final Checks


sananand007 (approved):
lgtm

PR #148, opened 2020-10-19
Add sign-up and login pages

weihesdlegend:
## Description
Create sign-up and log-in pages for access control.


templates/signup_page.html
```diff
@@ -0,0 +1,104 @@
...
+                <div class="col-md-4 mb-3">
+                    <label for="password"> Password </label>
+                    <input id="password" name="password" type="password" placeholder="Password" class="form-control"
+                           required>
+                </div>
+            </div>
+            <div class="form-row">
+                <div class="col-md-4 mb-3">
+                    <label for="email"> Email </label>
+                    <input id="email" name="email" type="email" placeholder="Johnny_jay@gmail.com" class="form-control"
```
sananand007, line 46: @weihesdlegend can you put the placeholder as just "Email" instead?

sananand007 (commented):
@weihesdlegend I dont think we need a username and a email ? We can just use the email to sign-in ? Basically only 2 blocks . It looks cleaner ?

weihesdlegend:
> @weihesdlegend I dont think we need a username and a email ? We can just use the email to sign-in ? Basically only 2 blocks . It looks cleaner ? 

The backend does not assume username equals email. But that is the practice for many websites. We should discuss if that is a better way. Currently without changing backend we do need all the 3 fields when they sign up.

PR #153, opened 2020-11-04
Use batch size to control number of places to update in each migration call

weihesdlegend:
## Description
To reduce multi-thread burden of the migration methods

## Root Cause
Previous method does not work properly on production

## Solution
use an extra parameter `batchSize` to control the maximum number of places to migrate for each call.
report how many places left to migrate so caller know if we still need to call the endpoint

## Testing
manual integration tests

## Final Checks

### DIFF
--- a/iowrappers/data_migrations.go
+++ b/iowrappers/data_migrations.go
@@ -2,16 +2,20 @@ package iowrappers
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


xtpitt (approved):
Nice job. Enjoy testing withya.

PR #158, opened 2020-11-09
modify Dockerfile to fix go build issue

sananand007:
## Description
We do not need to do a `cd main` any more as the main.go is in the working directory

## Root Cause
There was no `main` directory which was causing the `go build` step to fail in the Docker build step

## Solution
Changed the `Dockerfil` to take out the `cd main` 

## Testing
NR

## Final Checks
- [X] Have you removed commented code ?


Dockerfile
```diff
@@ -1,10 +1,9 @@
 FROM golang:latest as unwindtest
-RUN ls -altr
 
 RUN ls -al
```
weihesdlegend, line 4: Is this needed?

Dockerfile
```diff
@@ -1,10 +1,9 @@
...
 COPY . /app/
 WORKDIR /app
 
 RUN go get -v -t -d ./...
-RUN ls -altr && cd main/ && go build -v .
+RUN ls -altr &&  go build -v .
```
weihesdlegend, line 8: Go build -v . Should be sufficient

weihesdlegend (changes requested):

sananand007:
@weihesdlegend Are all changes in ? Could we approve

weihesdlegend:
> @weihesdlegend Are all changes in ? Could we approve

No, where is go test? Do we want to have run server part also?

sananand007:
Go test is a separate test, look at the build logs. Do we need to have the test run the docker image? Any reasons for that

weihesdlegend:
> Go test is a separate test, look at the build logs. Do we need to have the tests run within the docker image? Any reasons for that
> 
> ```
> function run_docker() {
> 	docker run --rm unwindtest:$BUILD_NUMBER /bin/sh -c "echo $PWD && \
> 		go test -v ./..."
> }
> ```

That's fair, but can you try to make it run under some port with `CMD`?

.gitignore
```diff
@@ -5,3 +5,4 @@ bin/
 .idea
 .DS_Store
 lock.json
+planner-app/
```
weihesdlegend, line 8: remove this line since it does not exist on master

weihesdlegend (approved):
LGTM

PR #167, opened 2021-05-22
Add Docker environment to support test development on Digital ocean droplets

sananand007:
## Description
This is an enhancement where we can host our GO server on DO droplets 

## Root Cause
This PR is for getting a test environment  ready to be used within a Docker container

## Solution

## Testing
No tests were added, and might be added on review

## Final Checks
- [X] Have you removed commented code ?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,15 @@
 FROM golang:alpine as unwindtest
 
 ENV GO111MODULE=on \
-    CGO_ENABLED=0
+    CGO_ENABLED=0 \
+    GOFLAGS=-mod=mod
 
 COPY . /app/
 WORKDIR /app/
 
 RUN ls -al
 
+EXPOSE 8010
+RUN go install -v github.com/weihesdlegend/Vacation-planner
 RUN go build -v .
+CMD ["/go/bin/Vacation-planner"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,11 +1,15 @@
...
+    GOFLAGS=-mod=mod
 
 COPY . /app/
 WORKDIR /app/
 
 RUN ls -al
 
+EXPOSE 8010
+RUN go install -v github.com/weihesdlegend/Vacation-planner
 RUN go build -v .
```
weihesdlegend, line 14: Can we try to remove build if install already covers it?

planner/planner_API.go
```diff
@@ -441,7 +441,7 @@ func (planner MyPlanner) SetupRouter(serverPort string) *http.Server {
 	}
 
 	svr := &http.Server{
-		Addr:         ":" + serverPort,
+		Addr:         "0.0.0.0:" + serverPort,
```
weihesdlegend, line 444: Would this work for Heroku?

Dockerfile
sananand007: yeah sure.. I will try that and check

planner/planner_API.go
sananand007: it shoud, do you think we can just deploy the branch and try out

weihesdlegend: I deployed a testing app: https://dashboard.heroku.com/apps/vacation-pla-feature-ad-epm8rp

PR #168, opened 2021-05-23
Feature: Use Bootstrap collapse for the plans page

weihesdlegend:
## Description
Use collapse to decorate the planning result page.

## Root Cause
Make the planning result page look more clean.

## Solution
Bootstrap CSS and JS plugin: https://getbootstrap.com/docs/4.3/components/collapse/

### DIFF
--- a/templates/common_style.css
+++ b/templates/common_style.css
@@ -2,48 +2,60 @@
     width: 400px;
     margin: 30px auto;
 }
+
 .login-form form {
     margin-bottom: 15px;
     background: #f7f7f7;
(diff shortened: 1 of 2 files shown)


sananand007 (approved):
lgtm,
I think at this stage it is good to have this, but we can keep this on the roadmap 

- It looks repetitive to have the same button again, I think it needs to be cleaner, like in the previous page, itself we ask the user how many plans he would like to see, and based on that we can decide to show them all they need to see 
- Let's say the user says 3 Plans, we want the `1` button to Hide or Show the plans, and by default we keep them unhidden

PR #171, opened 2021-06-28
Fix: Remove travel plans with the same set of places

weihesdlegend:
Travel plans with the same set of places prevent users from seeing more places. They are merely a permutation of the same set of places with different time-slots. This commit removes those plans from consideration of the final set of plans displayed for the users.

## Description
See commit message.

## Root Cause
See commit message.

## Solution
Add a filter function to remove duplicated plans defined by the joint place IDs.

## Testing
Both.

## Final Checks


solution/generate_planning_solutions.go
```diff
@@ -92,6 +94,8 @@ func GenerateSolutions(context context.Context, timeMatcher *matching.TimeMatche
 		mdIter.Next()
 	}
 
+	solutions = TravelPlansDeduplication(solutions)
```
sananand007, line 97: Nice 👍

solution/generate_planning_solutions.go
```diff
@@ -116,6 +120,23 @@ func GenerateSolutions(context context.Context, timeMatcher *matching.TimeMatche
...
 }
 
+//TravelPlansDeduplication removes travel plans contain places that are permutations of each other
+func TravelPlansDeduplication(travelPlans []PlanningSolution) []PlanningSolution {
+	duplicatedPlans := make(map[string]bool)
+	results := make([]PlanningSolution, 0)
+
+	for _, travelPlan := range travelPlans {
+		placeIds := travelPlan.PlaceIDS
+		radix.Sort(placeIds)
```
sananand007, line 130: any specific reason to have radix sort? I believe it's O(n) in TC, but just wanted to understand why choose this ?

sananand007 (approved):
lgtm

solution/generate_planning_solutions.go
weihesdlegend: Yes, for our case the N is small. But generally we want the linear time complexity.

PR #173, opened 2021-07-06
Display username on login success

weihesdlegend:
## Description
Most changes are reformatting. Please focus on the real changes.

## Root Cause
When users log in, they should see their usernames.

## Solution
login page passes username to the index page

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -1,171 +1,191 @@
 <!DOCTYPE html>
 <html lang="en">
+
 <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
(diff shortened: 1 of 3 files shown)


sananand007:
looks good. 

Could you make it the `same` color as `About` on the Left-hand corner?

sananand007 (approved):

sananand007 (changes requested):
as discussed seems this still needs changes

weihesdlegend:
> looks good.
> 
> Could you make it the `same` color as `About` on the Left-hand corner?

The About is a link, and currently we don't have user profile, so this will be just a `span` tag for text content.

sananand007 (approved):
lg

PR #174, opened 2021-07-10
Fix: Search page receives cookies and verifies them

weihesdlegend:
## Description
Currently after logged-in user goes to search page, the username is lost. This is confusing since users would not know if they logged in or not.

## Root Cause
Previous implementation has flaws. Cookies are only available on /v1 path endpoints.

## Solution
- Redirect home page request to /v1
- Search page verifies JWT token and use username from cookies to render
- Added JS-cookie library can be used for logging out users as well

## Final Checks
- [ x] You have added unit tests

### DIFF
--- a/templates/login_page.html
+++ b/templates/login_page.html
@@ -77,7 +77,7 @@
                 XHR.onload = function () {
                     if (XHR.readyState === XHR.DONE) {
                         if (XHR.status === 200) {
-                            window.location = "/?username=" + username;
+                            window.location = "/";
                         } else if (XHR.status === 401) {
                             alert("login failed, please check your credentials");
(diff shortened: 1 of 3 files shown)


templates/login_page.html
```diff
@@ -77,7 +77,7 @@ <h2 class="text-center"></h2>
                 XHR.onload = function () {
                     if (XHR.readyState === XHR.DONE) {
                         if (XHR.status === 200) {
-                            window.location = "/?username=" + username;
+                            window.location = "/";
```
sananand007, line 80: nice 👍

sananand007 (approved):
lgtm

PR #177, opened 2021-09-12
Migrate search results page to Bootstrap 5 with flush accordions

weihesdlegend:
## Description
Migrate search results page to Bootstrap 5 with flush accordions

## Root Cause
First PR to migrate HTML pages to use Bootstrap 5. Bootstrap 5 drops support for jQuery, etc.

## Solution
With flush accordions, each accordion opening folds the rest accordions.


rwangsc18 (approved):
Approved!

PR #178, opened 2021-09-18
Feature: Give search page a cleaner look

weihesdlegend:
## Description
The search page is due for a major revamp. The page used to have too many forms for users to fill, including the distance to city input which no one bothers to fill. The new design lets users select date and type in location information.

## Solution
* Use Bootstrap 5 components. 
* Autofill function is also updated to accommodate the new UI.
* Location autocomplete is not available.


templates/search_page.html
```diff
@@ -36,22 +37,37 @@ <h1 style="color: #24c1e0; text-align: center; background-color: whitesmoke">
     Vacation Planner
 </h1>
 
+<!--forms and inputs-->
 <div class="container" style="align-items: center">
```
rwangsc18, line 41: May consider adding some margin between this div.container and the above h1 header. 
For example, `<div class="container mt-3" style="align-items: center"> `
`mt-3` adds margin on the top of this div.

assets/js/search.js
```diff
@@ -40,10 +40,8 @@ function locateMe() {
                     response.json().then
```
rwangsc18, line 40: We may also use a good library called axios to simplify fetch API, the syntax will be something like 
`const data = await axios.get(url[, config])`
https://github.com/axios/axios#request-method-aliases

rwangsc18 (commented):
The new search page design looks good overall. I have also tested autofill functionality, and it works although sometimes the address is slow to show.
I have also added two comments.

templates/search_page.html
weihesdlegend: Done.

assets/js/search.js
weihesdlegend: Done. Great suggestion 💯

rwangsc18 (approved):

PR #185, opened 2021-10-03
Bug fix of missing date display in search page after clicking autofill

rwangsc18:
## Description
Discover that the main search page is missing date display after clicking autofill button.

## Root Cause
The date value needs comply with the format yyyy-mm-dd for the date input. For certain dates such as 10/01/2021, the day value can be a single digit (1) which violates the requirement.

## Solution
Prepend a "0" to day, if day is less than 10.

## Testing
Manual integration tests locally for functionalities including autofill, trip result searching etc.

## Final Checks

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -32,6 +32,7 @@ function locateMe() {
         const today = new Date();
 
         console.log(latitude, longitude);
+        console.log(today);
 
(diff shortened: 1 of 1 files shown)


weihesdlegend (approved):
Looks great 👍

PR #186, opened 2021-10-03
Update Dockerfile for the service to launch

weihesdlegend:
## Description
Update Dockerfile for the service to launch

## Solution
Expose port 10000 as usual


sananand007:
lgtm

sananand007 (approved):
It would be nice to check later within DO, I can try that later

PR #187, opened 2021-10-04
Fix: Set only JWT in session cookies

weihesdlegend:
## Description
Previously, after users logged in, the backend server assigns both JWT and a username to the session cookies. The username cookie is redundant and confusing to the front-end as well as backend functions. In this change, we attempt to remove the username from the cookie assignments and extract the username from the JWT claims.

## Solution
* Use `jwt-decode` module to extract claims on the front-end
* Simplify backend user authentication methods

### DIFF
--- a/assets/js/jwt-decode.js
+++ b/assets/js/jwt-decode.js
@@ -1,0 +1,116 @@
+/**
+ * The code was extracted from and modified based on:
+ * https://github.com/auth0/jwt-decode/blob/222db61fbaeea8ffd412a306039fea769ce43093/build/jwt-decode.js
+ */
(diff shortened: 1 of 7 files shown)


rwangsc18 (approved):
Approve.
I still need more time to fully understand the changes. But the idea to use JWT instead of session cookies to extract information is clear to me.

sananand007:
I am still looking at it @weihesdlegend

weihesdlegend:
> I am still looking at it @weihesdlegend 

Sure, feel free to comment away. Also see the deployment if you want test it.

sananand007 (approved):
lg.. very nice funtionality

PR #190, opened 2021-10-07
Fix: trip plan can have some times have less than 3 places

rwangsc18:
## Description
We sometimes see trip search results with trip that has only 2 places. 
For example, we search "10/06/2021" and "Newark, United States". 

## Root Cause
Early return in CreatePlanningSolutionCandidate() when 3rd and 1st locations are the same, which leads results of trip with 2 locations. 

## Solution
-- Add err to return values of CreatePlanningSolutionCandidate()
-- Candidate pools only consider returned plan without error

## Testing
Manual tested with date "10/06/2021" and location "Newark, United States", assuming previous search results are not cached. The return search page has all plans with 3 locations.

## Final Checks
- [Y] Have you removed commented code ?
- [Y] Have you used gofmt to format your code ?

### DIFF
--- a/solution/planning_solutions.go
+++ b/solution/planning_solutions.go
@@ -3,6 +3,7 @@ package solution
 import (
 	"container/heap"
 	"context"
+	"errors"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


rwangsc18:
> Nitpick: Use a single result object for both error and correct cases should be sufficient. The caller should always verify the error first.

Sure, I can try. Just to clarify, shall we add a new variable `Err` inside struct `PlanningSolution`? 
`type PlanningSolution struct {
	PlaceNames      []string            `json:"place_names"`
	PlaceIDS        []string            `json:"place_ids"`
	PlaceLocations  [][2]float64        `json:"place_locations"` // lat,lng
	PlaceAddresses  []string            `json:"place_addresses"`
	PlaceURLs       []string            `json:"place_urls"`
	PlaceCategories []POI.PlaceCategory `json:"place_categories"`
	Score           float64             `json:"score"`
	Err             error
}`

rwangsc18:
Final steps:

- squashed all related commits into one commit and rebase onto master
- performed some manual testing on this issue fix and new features such as user logout.

PR #191, opened 2021-10-13
Create user ID for new users

timwangmusic:
## Description
Previously we do not create a unique ID for new users. This change attempts to persist new users with a randomly generated UUID. Having a unique ID is essential for developing user customized features.

## Solution
* Use google UUID package
* When creating new users we create two records in Redis, with both username and ID postfixes.
* When query for existing users, add option to query by ID.

## Testing

## Final Checks

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:alpine as unwindtest
+FROM golang:1.16-alpine
 
 ENV GO111MODULE=on
 
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/weihesdlegend/Vacation-planner
 
-go 1.15
+go 1.16
 
 require (
 	github.com/alicebob/miniredis/v2 v2.13.3


rwangsc18:
@weihesdlegend Could you help me understand why we need two structs User and View in user.go? They seem identical to me.. Are they used to support saving two copes of user profiles, as you mentioned in the PR description? One is probably for findByUserId and the other is for findByUsername? The overall changes look good to me.

weihesdlegend:
> @weihesdlegend Could you help me understand why we need two structs User and View in user.go? They seem identical to me.. Are they used to support saving two copes of user profiles, as you mentioned in the PR description? One is probably for findByUserId and the other is for findByUsername? The overall changes look good to me.

Yes, although they are identical, the View is supposed to be used by Front-End and the Front-End should only see the view; while the User struct is the Model of the MVC pattern where our internal representation in the back-end.

weihesdlegend:
> Approved. I am just curious why we need two copies of user in redis. My understanding is that two copies are to support different findBy methods (1. id 2. username)?

Yes that's the purpose. When new users sign up we need to find by username, while when applications need to find user they refer by ID.

PR #197, opened 2021-10-28
Feature: Basic travel plan editing

weihesdlegend:
## Description
Allows users to click the edit button and make changes to a travel plan and save the resulting plan to their profile. Disable edit button after save.

## Solution
* Add an Edit button for each plan to trigger table editing
* Use jQuery to read table before save


rwangsc18 (approved):
Approved! Tested with Redis Insight, the user plan saving and editing works.

PR #199, opened 2021-11-02
Fix: Add administration area level one to user input location

weihesdlegend:
## Description
When searching cities with same name, the planner always uses the most popular city. This is not ideal especially when using Autofill.

## Root Cause
Location handling mechanisms lacking `Administration area level one` info.

## Solution
Add `Administration area level one` info and allow users to enter either `City, Country` or `City, Region, Country` in the search bar.

## Final Checks

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -31,8 +31,6 @@ function updateUsername() {
     userProfileElement.innerText = username;
 }
 
-updateUsername();
-
 function logOut() {
     const cookieToRemove = "JWT";
     const jwt = Cookies.get(cookieToRemove);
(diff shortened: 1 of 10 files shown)


rwangsc18 (commented):
Shall we also update the default address loaded in the main search page?
Instead of "Los Angeles, United States", maybe we can have "Los Angeles, CA, United States". Users can be more encouraged to include admin area level info in the future search as well.

weihesdlegend:
> Shall we also update the default address loaded in the main search page? Instead of "Los Angeles, United States", maybe we can have "Los Angeles, CA, United States". Users can be more encouraged to include admin area level info in the future search as well.

Done. Also I moved the set date function out and call it when page is loaded.

rwangsc18 (approved):
Approved.

PR #200, opened 2021-11-05
Fix: username field is undefined in the save plans method

weihesdlegend:
## Description
Save function is broken in search results page

## Root Cause
The new modular `updateUsername` method does not return values

## Testing


sananand007 (approved):
lg

rwangsc18 (commented):
LGTM

PR #207, opened 2021-11-14
Fix: Geocoding alias should trim empty location fields

weihesdlegend:
## Description
Geocoding alias should trim empty location fields

### DIFF
--- a/iowrappers/redis_client.go
+++ b/iowrappers/redis_client.go
@@ -268,46 +268,58 @@ func (redisClient *RedisClient) CacheLocationAlias(context context.Context, quer
 	return
 }
 
-// GetLocationWithAlias retrieves corrected location name from Redis; returns empty string if not exist;
-// corrects geocode query if corrected location name exists
-func (redisClient *RedisClient) GetLocationWithAlias(context context.Context, query *GeocodeQuery) string {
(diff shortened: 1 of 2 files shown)


iowrappers/redis_client.go
```diff
@@ -268,46 +268,58 @@ func (redisClient *RedisClient) CacheLocationAlias(context context.Context, quer
...
-func (redisClient *RedisClient) GetLocationWithAlias(context context.Context, query *GeocodeQuery) string {
+func (redisClient *RedisClient) GetLocationWithAlias(context context.Context, query *GeocodeQuery) (string, error) {
+	Logger.Debugf("(RedisClient)GetLocationWithAlias -> request: %+v", *query)
 	var err error
-	resCity, err := redisClient.client.HGet(context, "location_name_alias_mapping:city_names", strings.ToLower(query.City)).Result()
-	if err != nil {
-		return ""
+	var resCity, resAdminAreaLevelOne, resCountry string
+	var locationSegments []string
+	if strings.TrimSpace(query.City) != "" {
```
rwangsc18, line 276: So my interpretation about the new IF clause can catch any scenario that query string is either empty, or has multiple spaces. It makes sense to me.
` if strings.TrimSpace()!="" { } `

rwangsc18 (approved):
LGTM

PR #208, opened 2021-11-19
Data migration: remove places without URL or photo reference in database

weihesdlegend:
## Description
We noticed that a significant number of places data in production have incomplete fields, resulting in plans missing URLs or photo references. This deficiency is breaking our features such as trip details.

## Root Cause
When adding new fields such as photos, older places data become stale. Furthermore, some original data from Google lack those fields.

## Solution
* Remove places without URL or photo reference in database.
* Only admins can trigger the data migration.

## Final Checks

### DIFF
--- a/iowrappers/data_migrations.go
+++ b/iowrappers/data_migrations.go
@@ -2,6 +2,8 @@ package iowrappers
 
 import (
 	"context"
+	"fmt"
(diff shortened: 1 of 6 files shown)


rwangsc18 (approved):
approved

PR #212, opened 2021-12-21
Refactor: Add type to price level

weihesdlegend:
## Description
Google Maps API provides price level from 0 to 4, previously the model is coupled with the range from the API. This change  attempts to reduce the coupling and ease further feature development.

## Final Checks

### DIFF
--- a/POI/places.go
+++ b/POI/places.go
@@ -46,7 +46,7 @@ type Place struct {
 	Address          Address        `bson:"address"`
 	FormattedAddress string         `bson:"formatted_address"`
 	Location         Location       `bson:"location"`
-	PriceLevel       int            `bson:"price_level"`
+	PriceLevel       PriceLevel     `bson:"price_level,string"`
 	Rating           float32        `bson:"rating"`
 	Hours            [7]string      `bson:"hours"`
 	URL              string         `bson:"url"`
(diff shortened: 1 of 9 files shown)


POI/places.go
```diff
@@ -46,7 +46,7 @@ type Place struct {
 	Address          Address        `bson:"address"`
 	FormattedAddress string         `bson:"formatted_address"`
 	Location         Location       `bson:"location"`
-	PriceLevel       int            `bson:"price_level"`
+	PriceLevel       PriceLevel     `bson:"price_level,string"`
```
rwangsc18, line 49: Curious what does the newly added ",string" mean here? I know that "price_level" is BSON/JSON tag

rwangsc18 (approved):
LGTM
I just posted one question about one of BSON tag change in `POI/places.go`.

POI/places.go
weihesdlegend: Good catch! Initially I wanted to make the level a string, removed it.

matching/pricing.go
```diff
@@ -1,27 +1,30 @@
...
-		return PriceLevel4
+	case POI.PriceLevelZero:
+		return PriceZeroMean
+	case POI.PriceLevelOne:
+		return PriceOneMean
+	case POI.PriceLevelTwo:
+		return PriceTwoMean
+	case POI.PriceLevelThree:
+		return PriceThreeMean
+	case POI.PriceLevelFour:
```
sananand007, line 25: Nice work. Looks very clean

sananand007 (commented):
looks good

PR #214, opened 2021-12-31
Feature: basic places pricing filtering

weihesdlegend:
## Description
Backend changes for the pricing feature. Note that the pricing is only on catering locations since Google API does not provide attraction pricing info.

## Solution
* Added a price range matcher in the matching layer
* Modified cache key for planning solutions

## Testing

## Final Checks

### DIFF
--- a/planner/planner_APIs.go
+++ b/planner/planner_APIs.go
@@ -462,7 +462,7 @@ func (planner *MyPlanner) getPlanningApi(ctx *gin.Context) {
 		if planningResp.StatusCode == solution.InvalidRequestLocation {
 			ctx.String(http.StatusBadRequest, err.Error())
 		} else if planningResp.StatusCode == solution.NoValidSolution {
-			errString := "No valid solution is found.\n Please try to search with larger radius."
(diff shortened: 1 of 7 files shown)


planner/planner_APIs.go
```diff
@@ -435,7 +439,10 @@ func (planner *MyPlanner) getPlanningApi(ctx *gin.Context) {
 	}
 	iowrappers.Logger.Debugf("[%s] The number of requested planning results is %s.", requestId, numResults)
 
-	planningReq := solution.GetStandardRequest(POI.Weekday(weekday), numResultsInt)
+	priceLevel := ctx.DefaultQuery("price", "2")
```
sananand007, line 442: So default level we always keep as `2` ?

sananand007 (approved):
so how does the backend takes the input from the frontend action ?

planner/planner_APIs.go
weihesdlegend: Yes, I kept the default with average price range value.

PR #215, opened 2022-01-15
[Price][Front End]Add select form option for price

sananand007:
## Description
Front end addition for the price selection feature

## Root Cause

## Solution

## Testing

## Final Checks

### DIFF
--- a/templates/search_page.html
+++ b/templates/search_page.html
@@ -120,13 +120,15 @@
         </div>
 
         <div class="row mt-1">
-            <form class="d-flex" action="/v1/plans" method="get">
-                <div class="input-group w-auto">
-                    <input class="form-control-sm border border-secondary" type="date" id="datepicker" name="date"
+            <form class="row d-flex flex-column flex-sm-row align-items-start align-items-sm-center pe-0"
+                action="/v1/plans" method="get">
+                <div class="col ps-1 pe-0 mt-1" id="dateDiv">
+                    <input class="form-control border border-secondary" type="date" id="datepicker" name="date"
                         value="2020-02-29">
                 </div>
 
-                <div class="container px-1 border border-secondary rounded-1 d-flex flex-row align-items-center">
+                <div class="container border border-secondary rounded-1 d-flex flex-row align-items-center col ms-1 mt-1"
+                    id="locationDiv">
                     <div class="fixedwidthicon">
                         <i id="searchIcon" class="bi bi-search mx-1 visible"></i>
                     </div>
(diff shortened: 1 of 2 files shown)


templates/search_page.html
```diff
@@ -139,7 +140,29 @@ <h1 style="color: #24c1e0; text-align: center; background-color: whitesmoke">
...
+                    <select name="price" class="form-select border border-secondary"
+                        aria-label=".form-select-sm example" id="priceToSelect" onchange="getRandomInt(this)">
+                        <option>price</option>
+                        <option value="0">$</option>
+                        <option value="1">$$</option>
+                        <option value="2">$$$</option>
+                        <option value="3">$$$$</option>
+                        <option value="4">Max</option>
+                        <option value="0,1,2,3,4">Surprise</option>
+                        <script>
```
rwangsc18, line 153: How about putting JS logic codes (line 154 -- line 162) into `search.js`, because this HTML cites `search.js` anyway? So we can have a better organization of difference pieces of codes, html, css and JS etc.

rwangsc18 (commented):
Overall LGTM. Good job!
I also left one comment related with the new JS codes.

sananand007 (commented):
Can we merge t his now ? @weihesdlegend @rwangsc18

weihesdlegend:
> Can we merge t his now ? @weihesdlegend @rwangsc18 

I think we should discuss if vertical is a better solution

sananand007:
> > Can we merge t his now ? @weihesdlegend @rwangsc18
> 
> I think we should discuss if vertical is a better solution

Sure thing .. Let me also look into dynamic vertical solution on smaller screens ? that is what you are looking for rite ?

sananand007:
@weihesdlegend can you take a look, I used the `responsive` functionality on firefox, this looks roughly seamless, on change from 2 rows to 3, if this looks ok?
And the form inputs also work as usual

rwangsc18:
> Can we merge t his now ? @weihesdlegend @rwangsc18
@sananand007 
I think the new vertical design looks good to me. I left one comment about JS codes, and if it makes sense to place them inside JS script. Could you also check that one?

templates/search_page.html
rwangsc18: @sananand007 this is the comment that I shared earlier.

templates/search_page.html
```diff
@@ -140,9 +142,31 @@ <h1 style="color: #24c1e0; text-align: center; background-color: whitesmoke">
...
+                            function getRandomInt() {
```
weihesdlegend, line 156: Can we rename this to randomPriceRange? Try avoid using get in names unless it's http handling method with Get

weihesdlegend (approved):

PR #217, opened 2022-01-24
[Feature]Planning customization

weihesdlegend:
## Description
Provide a mechanism for users to customize plans at a refined level

## Root Cause
Towards our mission of providing more user customizations.

## Solution

## Testing

## Final Checks

### DIFF
--- a/assets/js/plan_template.js
+++ b/assets/js/plan_template.js
@@ -89,7 +89,8 @@ function tableToSlots() {
 async function postPlanTemplate() {
     document.getElementById("searchSpinner").classList.remove("visually-hidden");
     const location = document.getElementById('location').value.toString();
-    const locationFields = location.split(", ");
+    const locationFields = location.split(",");
+    locationFields.map(field => field.trim());
     let locationToPost = {}
     switch (locationFields.length) {
         case 2:
--- a/assets/js/plan_template.js
+++ b/assets/js/plan_template.js
@@ -27,11 +27,9 @@ function insertNewRow(start = '8', end = '10', category = 'Visit') {
     let attractionOption = document.createElement("option");
 
     let startTime = hourDropdown();
-    startTime.classList.add("form-select");
     startTime.value = start;
 
     let endTime = hourDropdown();
-    endTime.classList.add("form-select");
     endTime.value = end;
 
     cafeOption.text = "Eatery";
@@ -43,14 +41,13 @@ function insertNewRow(start = '8', end = '10', category = 'Visit') {
     select.value = category;
 
     newCell.append(select);
-
     newCell2.append(startTime);
-
     newCell3.append(endTime);
 }
 
 function hourDropdown() {
     let select = document.createElement("select");
+    select.classList.add("form-select");
     for (let hour = 8; hour < 20; hour++) {
         let option = document.createElement("option");
         option.text = hour.toString();
@@ -69,24 +66,26 @@ function tableToJSON() {
     )
 }
 
+function rowToSlot(row) {
+    return {
+        "category": row["Category"],
+        "time_slot": {
+            "slot": {
+                "start": parseInt(row["Start"]),
+                "end": parseInt(row["End"])
+            }
+        }
+    }
+}
+
 function tableToSlots() {
     const rows = tableToJSON()
     return rows.map(
-        function (row) {
-            return {
-                "category": row["Category"],
-                "time_slot": {
-                    "slot": {
-                        "start": parseInt(row["Start"]),
-                        "end": parseInt(row["End"])
-                    }
-                }
-            }
-        }
+        row => rowToSlot(row)
     )
 }
 
-async function postTemplate() {
+async function postPlanTemplate() {
     const location = document.getElementById('location').value.toString();
     const locationFields = location.split(",");
     let locationToPost = {}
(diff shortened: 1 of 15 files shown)


templates/plan_template.html
```diff
@@ -0,0 +1,159 @@
...
+
+                </table>
+            </div>
+        </div>
+
+        <div class="row">
+            <div class="btn-group-sm mb-3" role="group">
+                <button type="button" class="btn btn-outline-primary" id="add-row">add</button>
+                <button type="button" class="btn btn-outline-primary" id="remove-row">remove</button>
+                <button type="button" class="btn btn-outline-success" id="submit"> Search</button>
```
rwangsc18, line 129: How about aligning the value of `id` with its name? Maybe we can change the name from "Search" to "search", all lowercase, so it would be consistent with other buttons

assets/js/plan_template.js
```diff
@@ -0,0 +1,175 @@
...
+}
+
+function insertNewRow(start = '8', end = '10', category = 'Visit') {
+    const template = document.getElementById("template");
+    let newRow = template.insertRow(-1);
+    let newCell = newRow.insertCell(0);
+    let newCell2 = newRow.insertCell(1);
+    let newCell3 = newRow.insertCell(2);
+
+    let select = document.createElement("select");
```
rwangsc18, line 25: How about rename `select` as `categorySelect`?

assets/js/plan_template.js
```diff
@@ -0,0 +1,175 @@
...
+    await fetch(
+        url, {
+        method: "POST",
+        headers: {
+            'Content-Type': 'application/json'
+        },
+        body: JSON.stringify(data)
+    }
+    ).then(response => response.json())
+        .then(
```
rwangsc18, line 122: Are we using this promise chain with a series of `then()` in JS? https://javascript.info/promise-chaining
If so, how about use a different variable name such as `data` instead of `response` in line 123?

Line 123 could be `data => parseResponse(data)`.

assets/js/plan_template.js
```diff
@@ -0,0 +1,175 @@
...
+            }
+        }
+    )
+}
+
+function tableToSlots() {
+    const rows = tableToJSON()
+    return rows.map(
+        function (row) {
+            return {
```
rwangsc18, line 76: Maybe we could consider defining a new struct here to simplify codes between line 76 and 85. 
For example, we could define a struct or class in JS `Slot = {"attr": val, ...}`, a slot constructor `makeSlot(row)`.

rwangsc18: Then the return of tableToSlots() can be simplified into 
`return rows.map( row => makeSlot(row) )`

assets/js/plan_template.js
```diff
@@ -0,0 +1,175 @@
...
+    let select = document.createElement("select");
+    let cafeOption = document.createElement("option");
+    let attractionOption = document.createElement("option");
+
+    let startTime = hourDropdown();
+    startTime.classList.add("form-select");
+    startTime.value = start;
+
+    let endTime = hourDropdown();
+    endTime.classList.add("form-select");
```
rwangsc18, line 34: Could we also include modifying classList of `endTime` inside the function `hourDropdown()`, similar for the value attr?

assets/js/plan_template.js
```diff
@@ -0,0 +1,175 @@
...
+    cafeOption.text = "Eatery";
+    attractionOption.text = "Visit";
+
+    select.classList.add("form-select");
+    select.add(cafeOption);
+    select.add(attractionOption);
+    select.value = category;
+
+    newCell.append(select);
+
```
rwangsc18, line 46: Redundant empty line break?

rwangsc18 (commented):
Great job! I have also tried the customized search with more than 3 entries, and it works like a charm :)

I have left some comments or suggestions, and most of them are on JS syntax. Take a look and let me what you think.

assets/js/plan_template.js
weihesdlegend: Done.

assets/js/plan_template.js
weihesdlegend: We should always use `axios`! 👍🏼

weihesdlegend (commented):
@rwangsc18 Thanks for reviewing! Addressed the comments.

templates/plan_template.html
```diff
@@ -0,0 +1,167 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>Template</title>
+    <!--Bootstrap CSS-->
+    <link rel="stylesheet" href="assets/css/styles.css">
```
rwangsc18, line 9: If we place the link of our self-defined styles.css after bootstrap.css, we should have the proper height of banner.
For reference you can check the order of <link> in search_page.html.

weihesdlegend: Done

assets/js/plan_template.js
```diff
@@ -110,21 +109,16 @@ async function postTemplate() {
...
-        body: JSON.stringify(data)
-    }
-    ).then(response => response.json())
-        .then(
-            response => parseResponse(response)
-        )
-        .catch(
-            err => console.error(err)
-        )
+    axios.post(
```
rwangsc18, line 115: I see. axios does look a lot simpler

assets/js/plan_template.js
```diff
@@ -89,14 +89,15 @@ function tableToSlots() {
 async function postPlanTemplate() {
     document.getElementById("searchSpinner").classList.remove("visually-hidden");
     const location = document.getElementById('location').value.toString();
-    const locationFields = location.split(",");
+    const locationFields = location.split(", ");
```
rwangsc18, line 92: I see why we use ", " instead of "," as the separation symbol, because it takes care of the extra spaces in elements of `locationFields`.
Maybe we could use `locationFields = location.split(",").trim()`, because it can work with more cases?

rwangsc18 (approved):
Approved. I just left one more comment on the new changes.
We should be ready to merge this PR afterwards.

assets/js/plan_template.js
weihesdlegend: Great suggestion, but we cannot directly call trim after `split`. Added a `map` for trimming the fields

PR #218, opened 2022-01-30
Fix: Display error message if no valid plan can be found

weihesdlegend:
## Description

## Root Cause

## Solution

## Testing

## Final Checks

### DIFF
--- a/assets/js/plan_template.js
+++ b/assets/js/plan_template.js
@@ -119,7 +119,8 @@ async function postPlanTemplate() {
         function (response) {
             console.log(response.data);
             parseResponse(response.data);
-            document.getElementById("searchSpinner").classList.add("visually-hidden");
+            $('#searchSpinner').addClass("visually-hidden");
+            $('#no-valid-plan-error-msg').addClass('d-none');
         }
     ).catch(
         err => console.error(err)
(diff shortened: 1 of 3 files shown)


assets/js/plan_template.js
```diff
@@ -149,8 +150,9 @@ async function parseResponse(response) {
     console.log("Raw JSON response is", response);
 
     $(function () {
-        if (response["travel_plans"].length > 0) {
-            document.getElementById("results").classList.remove("d-none");
+        const resultsTable = document.getElementById("results");
+        if (response["travel_plans"]?.length > 0) {
```
rwangsc18, line 154: Curious what does "?" do here in if ().

weihesdlegend: Prevent the variable before `?` equals `null`

rwangsc18: Cool, learned something new! I have also searched a bit, and it is called optional chaining in JS guide. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

templates/plan_template.html
```diff
@@ -149,6 +149,10 @@ <h1 style="color: #24c1e0; text-align: center; background-color: whitesmoke">
             <tbody id="results-table-body">
             </tbody>
         </table>
+
+        <div class="d-flex d-none" id="no-valid-plan-error-msg">
```
rwangsc18, line 153: We may also consider using `alert` class in Bootstrap. It could help give consistent alert msg styles across different places, in case we add new ones in the future. 
https://getbootstrap.com/docs/5.0/components/alerts/

rwangsc18 (approved):
LGTM

templates/plan_template.html
weihesdlegend: Updated to use the alerts, looks so much better!

PR #220, opened 2022-02-06
[Enhancement]Filter places business status operational

weihesdlegend:
## Description
Due to recent events such as COVID-19, many business are affected. As such place data from Google Maps tend to change relatively more frequently. Therefore it is important we filter out business that are temporarily closed or permanently closed.

## Root Cause
Planning results contain temporarily closed or permanently closed places.

## Solution
Filter non-operational places in the search client.

## Final Checks

### DIFF
--- a/test/redis_client_mocks/nearby_search_test.go
+++ b/test/redis_client_mocks/nearby_search_test.go
@@ -8,7 +8,7 @@ import (
 
 func TestGetPlaces(t *testing.T) {
 	// set up data
-	places := make([]POI.Place, 3)
+	places := make([]POI.Place, 4)
 	places[0] = POI.Place{
 		ID:               "1001",
 		Name:             "Empire state building",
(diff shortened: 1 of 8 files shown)


test/redis_client_mocks/nearby_search_test.go
```diff
@@ -60,10 +76,11 @@ func TestGetPlaces(t *testing.T) {
 	// test normal cases
 	nycLatLng := POI.Location{Longitude: -74.006000, Latitude: 40.712800}
 	placeSearchRequest := iowrappers.PlaceSearchRequest{
```
rwangsc18, line 78: Great to see you have added more test cases. It seems that we are testing with multiple scenarios in this function, because there are different requests? Unless tests have dependencies with each other, it probably makes the future maintenance life easier if we split them into smaller tests.

rwangsc18 (commented):
Changes look good to me. 
Left one comment regarding with the new test case.

test/redis_client_mocks/nearby_search_test.go
weihesdlegend: Done

rwangsc18 (approved):
Approved

PR #222, opened 2022-02-08
Adjust styles of plan customization result tables 

weihesdlegend:
## Description
Align look and feel of plan customization result tables with those in the search results page


rwangsc18 (approved):
LGTM

PR #224, opened 2022-02-25
Preload city list when server starts

weihesdlegend:
## Description
Reduce backend calls to get city list when user types in the location

## Solution
Load city list when server starts

## Testing

## Final Checks

### DIFF
--- a/planner/planner_APIs.go
+++ b/planner/planner_APIs.go
@@ -30,6 +30,8 @@ const (
 	PhotoApiBaseURL    = "https://maps.googleapis.com/maps/api/place/photo?maxwidth=400&photo_reference=%s&key=%s"
 )
 
+var geocodes map[string]string
+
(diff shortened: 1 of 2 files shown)


planner/planner_APIs.go
```diff
@@ -30,6 +30,8 @@ const (
 	PhotoApiBaseURL    = "https://maps.googleapis.com/maps/api/place/photo?maxwidth=400&photo_reference=%s&key=%s"
 )
 
+var geocodes map[string]string
```
rwangsc18, line 33: I am curious what is the variable scope of `geocodes`. Is it a global variable for server application?

rwangsc18 (approved):
LGTM

planner/planner_APIs.go
weihesdlegend: It is a global variable.

PR #225, opened 2022-02-26
Fix the issue of seeing spinner when going back to the search page

rwangsc18:
## Description
The progress spinner was visible if we went back to the search page via hitting the back button on the browser. 
The spinner should be invisible in this case.

## Root Cause
Spinner DOM property was cached in the browser, and when the browser traverse back in history. The "visible" state is used for the spinner DOM.

## Solution
Add a new HTML event listener to hide the spinner whenever there is visibility change of the page.

## Testing
Manually tested with both fast loading and slow loading. Results are as expected.

## Final Checks

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -46,5 +46,10 @@ locationSearchInput.addEventListener(
         }
     }
 )
+// hide spinner when switching pages
+const hideSpinner = function () {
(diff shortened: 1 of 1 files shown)


weihesdlegend (approved):
LGTM

PR #228, opened 2022-03-06
Fix: Restore plan deduplication guarantee

weihesdlegend:
## Description

## Root Cause

## Solution

## Testing

## Final Checks


solution/planning_solutions.go
```diff
@@ -90,11 +91,15 @@ func FindBestPlanningSolutions(placeClusters [][]matching.Place, topSolutionsCou
 			log.Debug(err)
 			continue
 		}
+		if !plansDeduplication(deduplicatedPlans, candidate) {
```
rwangsc18, line 94: maybe we could rename the function as `isDuplicatePlan`
if isDuplicatePlan() {
    continue;
}

solution/planning_solutions.go
```diff
@@ -195,22 +200,22 @@ func GenerateSolutions(context context.Context, timeMatcher matching.Matcher, re
...
-		radix.Sort(placeIds)
-		jointPlanIds := strings.Join(placeIds, "_")
-		if _, exists := duplicatedPlans[jointPlanIds]; !exists {
-			results = append(results, travelPlan)
-			duplicatedPlans[jointPlanIds] = true
-		}
+// returns true if the new travel plan can be added to priority queue
+func plansDeduplication(plans map[string]bool, newPlan PlanningSolution) bool {
+	jointPlaceIDs := jointPlaceIDs(newPlan)
+	if val, exists := plans[jointPlaceIDs]; !exists || !val {
```
rwangsc18, line 206: Is `!exists` alone sufficient as the IF statement logic?

solution/planning_solutions.go
```diff
@@ -195,22 +200,22 @@ func GenerateSolutions(context context.Context, timeMatcher matching.Matcher, re
...
-		copy(placeIds, travelPlan.PlaceIDS)
-		radix.Sort(placeIds)
-		jointPlanIds := strings.Join(placeIds, "_")
-		if _, exists := duplicatedPlans[jointPlanIds]; !exists {
-			results = append(results, travelPlan)
-			duplicatedPlans[jointPlanIds] = true
-		}
+// returns true if the new travel plan can be added to priority queue
+func plansDeduplication(plans map[string]bool, newPlan PlanningSolution) bool {
+	jointPlaceIDs := jointPlaceIDs(newPlan)
```
rwangsc18, line 205: consider a different output var name, instead of jointPlaceIDs, maybe something like `planID`?
so we can also avoid reusing the same name with that of the function

rwangsc18 (commented):
LGTM overall.
I add some nitpicks on function/variable names.

rwangsc18 (commented):
LGTM overall
Left a few nitpicks on function/variable names.

rwangsc18 (commented):
LGTM overall. Leave a new nitpicks on function/variable names. plz take a look as well

solution/planning_solutions.go
weihesdlegend: Maybe we should remove keys from map.

solution/planning_solutions.go
weihesdlegend: Done.

solution/planning_solutions.go
weihesdlegend: Renamed to `jointPlaceIdsForPlan`.

solution/planning_solutions.go
rwangsc18: I agree

rwangsc18 (approved):
Approved

PR #231, opened 2022-03-22
Improve user authentications

weihesdlegend:
## Description
* Allows users to log in with username or email address
* Use Bootstrap toasts to replace plain JS alert when duplicated usernames are detected

## Solution
* Enhance Redis backend: created 2 hashes for mapping emails and usernames to user IDs

## Final Checks

### DIFF
--- a/templates/signup_page.html
+++ b/templates/signup_page.html
@@ -53,6 +53,18 @@
         </form>
     </div>
 
+    <div class="d-flex justify-content-center align-items-center">
+        <div class="toast" role="alert" aria-live="assertive" aria-atomic="true" id="sign-up-error-toast">
+            <div class="toast-header">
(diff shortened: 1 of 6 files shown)


templates/signup_page.html
```diff
@@ -53,6 +53,18 @@ <h1> Vacation Planner </h1>
         </form>
     </div>
 
+    <div class="d-flex justify-content-center align-items-center">
```
rwangsc18, line 56: nit: we can consider using warnings over alerts in Bootstrap.
https://getbootstrap.com/docs/5.0/components/alerts/

rwangsc18 (approved):
LGTM

templates/signup_page.html
weihesdlegend: Done, the only thing is user cannot remove the banner

PR #233, opened 2022-03-27
Feature: User log in with Google OAuth2

weihesdlegend:
## Description
* Basic Google OAuath2 implementation
* Users can use their Gmail email addresses to login

## Solution
* Depend on Golang OAuth package
* Users are still required to sign up in the sign-up page. If the user email is not signed up, they shall be redirected to the login page


rwangsc18 (approved):
LGTM

PR #234, opened 2022-04-15
Feature: use precise location for travel planning

weihesdlegend:
## Description
For metropolitan areas, using the current location often resolves to the downtown areas. We provide an option for users to use their exact current location for searching nearby places.

## Solution
Add a button in the search page for this new option.

The trickier part for the backend change is how to use cache properly. Redis is using lat/long for searching places. Therefore we need to feed the Redis client with precise location. At the same time, we need to maintain data freshness, therefore we need to reverse geocoding in order to find the city-level info in order to determine when to perform a new search.

## Testing

## Final Checks

### DIFF
--- a/assets/css/styles.css
+++ b/assets/css/styles.css
@@ -17,6 +17,9 @@
 div.fixedwidthicon {
     width: 25px;
 }
+button.square-button{
+    width: 31px;
+}
 
 h1 {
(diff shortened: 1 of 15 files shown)


rwangsc18 (approved):
LGTM

PR #235, opened 2022-04-23
Fix: Insensitive to user email for login and signup

weihesdlegend:
## Description
The authentication system do not treat user emails case-insensitive.

## Solution
* Save email addresses in lower case when users create accounts
* Check lower case converted email addresses when authenticating users.

Expected behaviors:
* When users log in with usernames, differentiate cases.
* When users log in with emails, do NOT differentiate cases.
* When users sign up they cannot use emails of same lower-cased form.

## Testing
- [ ] Added new unit tests

## Checks


rwangsc18 (approved):
LGTM

PR #236, opened 2022-04-28
[Fix] Correct the order of top plans returned from search

rwangsc18:
## Description
Bug was found when examining scores of top plans, and the current top plans are shown in a reverse order their plan scores. 

## Solution
The bug is due to the usage of min-heap. The result list is in ascending order instead of descending order. A simple fix is to reverse the output plan list.

## Testing
- [ ] Added new unit tests
- [x] Fixed broken unit tests

## Checks

### DIFF
--- a/solution/planning_solutions.go
+++ b/solution/planning_solutions.go
@@ -4,8 +4,9 @@ import (
 	"container/heap"
 	"context"
 	"errors"
-	log "github.com/sirupsen/logrus"
 	"strings"
+
+	log "github.com/sirupsen/logrus"
 
 	"github.com/google/uuid"
 	"github.com/yourbasic/radix"
@@ -113,8 +114,15 @@ func FindBestPlanningSolutions(placeClusters [][]matching.Place, topSolutionsCou
 		top := heap.Pop(priorityQueue).(graph.Vertex)
 		res = append(res, top.Object.(PlanningSolution))
 	}
+	// min-heap, res needs to be reversed to get the descending order
+	return reversePlans(res)
+}
 
-	return res
+func reversePlans(plans []PlanningSolution) []PlanningSolution {
+	for i, j := 0, len(plans)-1; i < j; i, j = i+1, j-1 {
+		plans[i], plans[j] = plans[j], plans[i]
+	}
+	return plans
 }
 
 func GenerateSolutions(context context.Context, timeMatcher matching.Matcher, redisClient iowrappers.RedisClient, redisRequest iowrappers.PlanningSolutionsCacheRequest, request PlanningRequest, priceRangeMatcher matching.Matcher) (solutions []PlanningSolution, solutionRedisKey string, err error) {
--- a/test/solution_candidate_selection_test.go
+++ b/test/solution_candidate_selection_test.go
@@ -1,11 +1,12 @@
 package test
 
 import (
+	"strconv"
+	"testing"
+
 	"github.com/weihesdlegend/Vacation-planner/POI"
 	"github.com/weihesdlegend/Vacation-planner/matching"
 	"github.com/weihesdlegend/Vacation-planner/solution"
-	"strconv"
-	"testing"
 )
 
 // test if priority queue gives top solution candidates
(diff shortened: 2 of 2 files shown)


solution/planning_solutions.go
```diff
@@ -4,9 +4,10 @@ import (
 	"container/heap"
 	"context"
 	"errors"
-	log "github.com/sirupsen/logrus"
 	"strings"
 
+	log "github.com/sirupsen/logrus"
+
```
sananand007, line 10: intentional?

solution/planning_solutions.go
```diff
@@ -113,8 +114,15 @@ func FindBestPlanningSolutions(placeClusters [][]matching.Place, topSolutionsCou
 		top := heap.Pop(priorityQueue).(graph.Vertex)
 		res = append(res, top.Object.(PlanningSolution))
 	}
+	// min-heap, res needs to be reversed to get the descending order
+	return reverseArray(res)
```
sananand007, line 118: :+1:

test/solution_candidate_selection_test.go
```diff
@@ -1,11 +1,12 @@
 package test
 
 import (
+	"strconv"
```
sananand007, line 4: same, intentional?

sananand007 (approved):
lgtm, some [nit]s

solution/planning_solutions.go
```diff
@@ -113,8 +114,15 @@ func FindBestPlanningSolutions(placeClusters [][]matching.Place, topSolutionsCou
...
+}
 
-	return res
+func reverseArray(arr []PlanningSolution) []PlanningSolution {
```
weihesdlegend, line 121: Maybe put this function in utils and make it generic?

solution/planning_solutions.go
rwangsc18: This is corrected by Go code formatter.

solution/planning_solutions.go
rwangsc18: Good suggestion, will do so.

test/solution_candidate_selection_test.go
rwangsc18: Also corrected by Go code formatter

solution/planning_solutions.go
rwangsc18: Moved this function to solution/utils.go, please check e9be6c7.
Wanted to use the more generic style with abstract input data type, but it seems require higher Go version (1.18+) to support.

solution/utils.go
```diff
@@ -0,0 +1,8 @@
+package solution
+
+func ReverseSlice(arr []PlanningSolution) []PlanningSolution {
```
weihesdlegend, line 3: Maybe we can make this function generic and put it in the `utils` package? For example, `func ReverseSlice(arr []interface{}) []interface{} {...} `

rwangsc18: sg. I actually tried to use interface as arguments, but it somehow doesn't work as expected. I would get this error in VScode.
Screen Shot 2022-05-02 at 4 57 04 PM

rwangsc18: I have found there are some other examples, using the `T` type to indicate its generic data type. However it requires Go version newer than 1.18, and we are 1.16 I believe.
Screen Shot 2022-04-30 at 5 59 14 PM

weihesdlegend: We don't need generics here. Refer to this example I wrote: https://go.dev/play/p/aaW5z6Uc8GS

rwangsc18: I see, but is your example for `maps` not `slice` in golang? 
I have also found an article about interface slice, but it also suggests the following input and output don't work right away.
`func ReverseSlice(arr []interface{}) []interface{} {...}`

weihesdlegend: Makes sense, then we should just name the method ReversePlans and put it in the same file

sananand007 (approved):
lg

solution/utils.go
rwangsc18: Sure, I just updated the function in 21c6d64

PR #237, opened 2022-05-02
Fix: persist location input value when going back to the search page

weihesdlegend:
Also restore log-out function on the search page.

## Description
Setting location on change does not guarantee consistent location persistence behavior.

## Solution
* Set location in `SessionStorage` only when the form is submitted.
* Restore log out functionality.

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -1,6 +1,5 @@
 // methods for the search page
-import { locateMe, setDateToday, preciseLocation } from "./utils.js";
-
+import { locateMe, preciseLocation, setDateToday } from "./utils.js";
 import { logOut } from "./user.js";
 
 setDateToday();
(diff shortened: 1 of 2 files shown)


assets/js/search.js
```diff
@@ -30,17 +29,28 @@ setDateToday();
...
+const FORM = document.getElementById("main-form");
+const STORAGE_ITEM = "location";
+const LOCATION_INPUT = document.getElementById("location");
+$(document).ready(() => {
+    const val = sessionStorage.getItem(STORAGE_ITEM);
+    if (val) {
+        console.log(`Set the Location based on PageLoad...` + val);
+        LOCATION_INPUT.value = val;
+    }
 });
```
sananand007, line 46: @weihesdlegend so the function `fetchLocation` is being moved here? looks good, just need to make sure if that works

assets/js/search.js
```diff
@@ -30,17 +29,28 @@ setDateToday();
...
-    );
-    elem.value = sessionStorage.getItem("locationPerm");
-  }
+document.getElementById("logout-confirm-btn").addEventListener(
+    "click", logOut
+)
+
+const FORM = document.getElementById("main-form");
+const STORAGE_ITEM = "location";
+const LOCATION_INPUT = document.getElementById("location");
```
sananand007, line 38:
```suggestion
const LOCATION_INPUT = document.getElementById(STORAGE_ITEM);
```

assets/js/search.js
```diff
@@ -30,17 +29,28 @@ setDateToday();
...
-    console.log(
-      `Set the Location based on PageLoad...` +
-        sessionStorage.getItem("locationPerm")
-    );
-    elem.value = sessionStorage.getItem("locationPerm");
-  }
+document.getElementById("logout-confirm-btn").addEventListener(
+    "click", logOut
+)
+
```
sananand007, line 35: Do this part above was removed with my PR ?

sananand007 (approved):
lgtm... lets do some tests

assets/js/search.js
weihesdlegend: Yes

assets/js/search.js
weihesdlegend: Previously you used `onChange` to track modifications to the location input element. However it misses auto complete as well as find my location. So the change in this PR is a completely new approach.

assets/js/search.js
weihesdlegend: Will not adopt because the location happens to be the same name of the ID of the location input element.

assets/js/search.js
sananand007: by `find my` location, do you mean the arrow sign functionality that is on the right top of the price bar ? 
- I checked that is working on production on the web ?

assets/js/search.js
```diff
@@ -1,6 +1,5 @@
 // methods for the search page
-import { locateMe, setDateToday, preciseLocation } from "./utils.js";
-
+import { locateMe, preciseLocation, setDateToday } from "./utils.js";
```
rwangsc18, line 2: NIT: curious why the order of two functions are swapped? Is it done by the code formatter?

rwangsc18 (approved):
LGTM

assets/js/search.js
weihesdlegend: Yes, sorted alphabetically

PR #239, opened 2022-05-04
Fix: Redis plan search does not work with precise location

weihesdlegend:
## Description
Use precise location feature does not behave consistently. Only fresh searches return correct results. Some debugging work shows that the issue lies in Redis plan search uses city, country location, which is not available when using precise location feature.

## Solution
* Make sure to perform reverse geocoding before sending request to Redis

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/solution/solver.go
+++ b/solution/solver.go
@@ -88,10 +88,23 @@ func PlanningSolutionsRedisRequest(location POI.Location, placeCategories []POI.
 }
 
 func (solver *Solver) Solve(context context.Context, redisClient iowrappers.RedisClient, request *PlanningRequest, response *PlanningResponse) {
+	iowrappers.Logger.Debugf("->Solve(context.Context, iowrappers.RedisClient, %v, *PlanningResponse)", request)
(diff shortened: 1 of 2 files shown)


solution/solver.go
```diff
@@ -88,12 +88,24 @@ func PlanningSolutionsRedisRequest(location POI.Location, placeCategories []POI.
...
 		response.Err = errors.New("invalid travel destination")
 		response.ErrorCode = InvalidRequestLocation
 		return
 	}
 
+	if request.PreciseLocation {
+		geocode, err := solver.Searcher.ReverseGeocode(context, request.Location.Latitude, request.Location.Longitude)
+		if err != nil {
+			response.Err = err
+			response.ErrorCode = InvalidRequestLocation
```
rwangsc18, line 102: add `return` when the ReverseGeocode fails?

rwangsc18 (commented):
LGTM, left one comment

rwangsc18 (approved):
Approved!

PR #241, opened 2022-05-20
Enhancement: Send user verification emails

weihesdlegend:
## Description
Send email verification when users sign up.

## Solution
* Create a mailer for SendGrid integration
* Create temp user to hold user data while waiting for user clicking on the verification link

## Testing
- [ ] Added new unit tests

## Checks


iowrappers/users.go
```diff
@@ -85,7 +88,7 @@ func (redisClient *RedisClient) FindUser(context context.Context, findUserBy Fin
 	return userView, nil
 }
 
-func (redisClient *RedisClient) CreateUser(context context.Context, userView user.View) (user.View, error) {
+func (redisClient *RedisClient) CreateUser(context context.Context, userView user.View, skipPasswordGeneration bool) (user.View, error) {
```
rwangsc18, line 91: curious why we need `skipPasswordGeneration` with user email verification? Is it trying to skip password encryption? IIUC, the password is from user input, and we encrypted it before storing it on Redis. Does email verification process add some additional steps?

rwangsc18 (commented):
LGTM
left one comment, is the PR deployed somewhere, so we can test it?

iowrappers/users.go
weihesdlegend: > LGTM left one comment, is the PR deployed somewhere, so we can test it?

Deployed on https://staging-vp.herokuapp.com/v1/

weihesdlegend: > 

Yes, the purpose is to skip password encryption. When a user signs up, we cannot store the original password but the encrypted one temporarily. So the user creation function has to deal with both cases.

sananand007 (approved):
lg :t-rex:

PR #250, opened 2022-10-27
Refactor: replace Golang HTML template with profile page JS rendering

timwangmusic:
## Description
As titled.

## Testing
- [ ] Added new unit tests

## Checks


rwangsc18 (approved):
Approved.
This is actually great. Later we can use this type of similar template to make the page rendering more dynamic.

PR #251, opened 2022-10-29
Fix: user saved plan location format

timwangmusic:
## Description
When user auto completes the location with search input, the resulted string in search result page URL does not follow traditional formatting for city, country names, etc.

## Solution
Fix the format while posting the plan to be saved to the backend.

### DIFF
--- a/assets/js/search_results.js
+++ b/assets/js/search_results.js
@@ -1,6 +1,7 @@
-import {Place, View} from "./place.js";
+import { Place, View } from "./place.js";
 
-import {updateUsername} from "./user.js";
+import { capitalizeFirstChar } from "./utils.js"
+import { updateUsername } from "./user.js";
 
 let numberOfPlans = 5;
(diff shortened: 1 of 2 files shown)


assets/js/search_results.js
```diff
@@ -78,6 +81,19 @@ function planToView(plan, planIndex) {
     return view;
 }
 
+// The autocompleted locations need to be fixed. City and country names need to be capitalized and admin level 2 names need to be changed to all upper cases.
+function correctLocationName(location) {
```
rwangsc18, line 85: nit: call it normalizeLocation, since we just align the format to make things consistent

rwangsc18 (commented):
LGTM

assets/js/search_results.js
timwangmusic: Done

rwangsc18 (approved):
LGTM

PR #252, opened 2022-11-01
Feature: Pagination on plan template page

timwangmusic:
## Description
Pagination on plan template page.

## Solution
* Use the simple pagination plugin.
* The number of plans and number of items parameters are tentative and subject to change.

## Testing

## Checks

### DIFF
--- a/assets/js/plan_template.js
+++ b/assets/js/plan_template.js
@@ -4,6 +4,8 @@ import { locateMe, setDateToday } from "./utils.js";
 import { logOut, updateUsername } from "./user.js";
 
 const username = updateUsername();
+const totalItemsCount = 10;
+const itemsPerPage = 2;
 
 setDateToday();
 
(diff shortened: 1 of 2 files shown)


assets/js/plan_template.js
```diff
@@ -193,6 +195,22 @@ async function parseResponse(response) {
...
+
+                // Items allowed on a single page
+                itemsOnPage: itemsPerPage,
+                onPageClick: function (pageIdx) {
+                    const itemsOnPage = 2;
+                    $('#tables .table').hide()
+                        .slice(itemsOnPage * (pageIdx - 1),
+                            itemsOnPage + itemsOnPage * (pageIdx - 1)).show();
+                }
+            });
```
rwangsc18, line 213: This looks good to me. Just curious why do we need `parseResponse` to be async function? I forgot about this, is it something related ensuring the function execution order.

rwangsc18 (approved):
LGTM

assets/js/plan_template.js
timwangmusic: Good point, I think the function does not need to be async.

PR #253, opened 2022-11-06
Refactor: Plan generation timeout

timwangmusic:
## Description
We often see critical errors from deployed service that requests are timing out for more than 30 seconds. For large scale computations, currently we cannot generate a solution due to the complexity. The requests should not be blocked and should be timed out within reasonable time.

## Solution
* Use `context.WithTimeout` to limit `GenerateSolutions` run time.
* Use `timeout` Gin middleware to limit regular request run time.

## Testing
- [x] Added new unit tests

## Checks


sananand007 (approved):
lgtm

PR #254, opened 2022-11-08
Optimization: Sub-optimal solution generations

timwangmusic:
## Description
If we sort the places before feeding the iterator loop, we can just consider the top places and exit the loop as soon as there is enough number of solutions generated.

## Solution
* Sort places before running iterator loop and break loop as soon as we obtain enough planning solutions.

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/planner/solver.go
+++ b/planner/solver.go
@@ -11,6 +11,7 @@ import (
 	"github.com/weihesdlegend/Vacation-planner/iowrappers"
 	"github.com/weihesdlegend/Vacation-planner/matching"
 	"github.com/yourbasic/radix"
+	"sort"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


planner/solver.go
```diff
@@ -259,11 +260,13 @@ func FindBestPlanningSolutions(placeClusters [][]matching.Place, topSolutionsCou
 	priorityQueue := &graph.MinPriorityQueueVertex{}
 	deduplicatedPlans := make(map[string]bool)
 
+	for _, cluster := range placeClusters {
```
rwangsc18, line 263: is cluster equal to Place or []Place here? if it is the latter, the sorting complexity is not small right? because we are sorting all combinations of possible plans, which can be exponential?

rwangsc18 (commented):
left a comment related with the presorting part. The rest looks good to me.

planner/solver.go
timwangmusic: The cluster equals `[]Place`. The complexity is `O(NMlog(M))`, so it is not terrible. We are sorting individual lists not combinations.

rwangsc18: i see, the sorting is done on a list of resorts or restaurants. I assume N is the number of timeslots, and M is number of places available in that slot.

rwangsc18 (approved):
LGTM

sananand007 (approved):
lg

timwangmusic:
Closed because the plan quality is compromised.

PR #258, opened 2023-01-02
Feature: Add most searched place in user profile page

timwangmusic:
## Description
Enrich user profile page by adding a favorite section. This PR adds the most searched place by a user.

## Solution
* Add a new data structure `PersonalFavorites` to the user view.
* Count and record the timestamp when a user searches a place.
* Frontend renders the user favorites besides saved travel plans.

## Testing
- [x] Added new unit tests

## Checks


assets/js/user_profile.js
```diff
@@ -67,4 +79,24 @@ async function getUserPlans() {
...
+            const favorites = data['searchHistory'];
+            if (Object.keys(favorites).length > 0) {
+                renderFavorites(favorites);
+            }
+        }
+    ).catch(
+        err => console.error(err)
+    );
+}
+
```
sananand007, line 98: wondering what is the difference if we do 
```
async function getUserFavorites() {
	const favoritesUrl = `/v1/users/${username}/favorites`;
	try {
		response = await axios.get(favoritesUrl)
		const data = response.data;
        const favorites = data['searchHistory'];
        if (Object.keys(favorites).length > 0) {
                renderFavorites(favorites);
        }
	}
	catch (error) {
		console.error(`Err: ${err}`)
	}
}
```

sananand007 (approved):
lgtm

assets/js/user_profile.js
timwangmusic: `Axios` is implicitly doing the `try...catch...`.

rwangsc18 (commented):
LGTM. 
It is difficult to optimize HTML/CSS layout for the favorite box without a signed-in user, because the staging environment doesn't allow creating a user. If okay, let's update the layout in a separate PR after we merge this PR to prod.

PR #265, opened 2023-01-09
Refactor: use channel for FindBestPlanningSolutions function response

timwangmusic:
## Description
The `FindBestPlanningSolutions` method used to return `[]PlanningSolution`. The issue is when time out happens, the iterator's loop has to terminate and no meaningful result will be returned.

## Solution
* Refactor `FindBestPlanningSolutions` method to return `chan PlanningResp`.
* Refactor calling function code.
* Set context with timeout in `FindBestPlanningSolutions` parent function.

## Testing
- [x] Updated unit tests

## Checks

### DIFF
--- a/planner/solver.go
+++ b/planner/solver.go
@@ -12,7 +12,6 @@ import (
 	"github.com/weihesdlegend/Vacation-planner/matching"
 	"github.com/yourbasic/radix"
 	"strings"
-	"time"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


planner/solver.go
```diff
@@ -340,20 +337,24 @@ func (s *Solver) generateSolutions(ctx context.Context, req *PlanningReq, timeMa
...
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, SolverTimeout)
+	defer cancel()
+
+	select {
+	case <-ctxWithTimeout.Done():
+		resp.Err = errors.New("cannot complete computation in time")
 		resp.ErrorCode = RequestTimeOut
-		return resp, err
+		return
+	case r := <-FindBestPlanningSolutions(ctxWithTimeout, placeClusters, req.NumPlans, mdIter):
```
rwangsc18, line 355: maybe we can also use `resp` here? case resp <- FindBestPlanningSolutions()

planner/solver.go
```diff
@@ -340,20 +337,24 @@ func (s *Solver) generateSolutions(ctx context.Context, req *PlanningReq, timeMa
...
+		return r
```
rwangsc18, line 356: then we can simply write `return` here?

planner/solver.go
```diff
@@ -231,18 +224,20 @@ func createPlanningSolutionCandidate(placeIndexes []int, placeClusters [][]match
 	return res, nil
 }
 
-func FindBestPlanningSolutions(ctx context.Context, placeClusters [][]matching.Place, topSolutionsCount int, iterator *MultiDimIterator) ([]PlanningSolution, error) {
+func FindBestPlanningSolutions(ctx context.Context, placeClusters [][]matching.Place, topSolutionsCount int, iterator *MultiDimIterator) (resp chan PlanningResp) {
```
rwangsc18, line 227: we can explicitly specify that `resp` is a **send-only** channel in the function definition, so we prevent read operation in the function.

(resp chan<- PlanningResp)

rwangsc18 (approved):
LGTM

planner/solver.go
timwangmusic: if we do that then the caller `generateSolutions` cannot use the channel.

PR #272, opened 2023-01-30
FIx: logins via oauth do not show usernames

timwangmusic:
## Description
 Logins via OAuth do not show usernames.

## Solution
Because of the difficulty in setting up oauth in staging and test, we have to bump up logging levels to test in prod, which is bad practice.

## Checks


timwangmusic:
> LGTM. Thanks for the fix. Do you want to update the description of the commit, since it is not just about updating the logging level?

Done.

rwangsc18:
LGTM. Approved!

PR #275, opened 2023-02-04
Fix: persistent alert message when back forward or reload the page

timwangmusic:
## Description
The newly added alert message when no plan is not found persist when users swipe backward on phones or reload the page.

## Solution
* This change fixes the issue by making sure the navigation is of `navigate` type instead of the other types described [here](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming/type).

## Testing

## Checks


rwangsc18 (approved):
LGTM

PR #287, opened 2023-03-16
Verify password reset requests with code

timwangmusic:
## Description
In the previous PR, we already created a verification code for each password reset request. But we did not make proper use of it. Also users should not need old password to set new ones, otherwise why would they need to reset?

## Solution
* Verify user email against verification code to make sure URL is not tampered.
* Verify password reset code is claimed only once.
* Removed requirement for user entering old password.

## Testing

## Checks


rwangsc18 (approved):
LGTM

PR #289, opened 2023-04-14
Fix: precise location flag resets when user is navigated

timwangmusic:
## Description
When users are navigated back to the search page, they will still see the location input unchanged since we persist this in session storage. However when users attempt to search again without changing the input, they will see the error message `location format must be city, region, country or city, country`. This is cause by `precise-location-flag` being reset to `false` upon loading the page.

## Solution
* Persist `precise-location-flag` value in session storage
* Load the persisted value upon page loading
* Reset `precise-location-flag` value when user intentionally changes input

## Testing
- [ ] Added new unit tests

## Checks


rwangsc18 (approved):
LGTM

PR #296, opened 2023-06-06
Feat: solve optimal planning problem with Hungarian Algorithm

timwangmusic:
## Description
We can model the planning problem in a Bipartite graph where the edges connect two sets with a set of time slots and a set of places. Then the problem reduces to find a maximum matching in the graph. The weights of the edges are scores of the places. We can use a Hungarian Algorithm solver to solve the maximization problem.

## Solution
* Compute a weight matrix for the Bipartite graph.
* Fix weight so we can solve a variation of the minimization problem which the Hungarian Algorithm designed to solve.
* Return place IDs as the solution.
* Example request: `{
    "location": {
        "city": "Shanghai",
        "country": "China"
    },
    "slots": [
        {
            "time_slot": {
                "slot": {
                    "start": 10,
                    "end": 12
                }
            },
            "category": "Visit"
        },
        {
            "time_slot": {
                "slot": {
                    "start": 14,
                    "end": 16
                }
            },
            "category": "Visit"
        }
    ],
    "radius": 10000,
    "weekday": 5,
    "price_level": 2
}`

## Testing
- [x] Added new unit tests

## Checks

### DIFF
--- a/planner/planner.go
+++ b/planner/planner.go
@@ -349,6 +349,20 @@ func (p *MyPlanner) homePageHandler(ctx *gin.Context) {
 	ctx.Redirect(http.StatusMovedPermanently, "/v1/")
 }
 
+func (p *MyPlanner) getOptimalPlan(ctx *gin.Context) {
+	req := &PlanningReq{}
+
+	if err := ctx.ShouldBindJSON(req); err != nil {
+		ctx.JSON(http.StatusBadRequest, gin.H{"error": err})
+	}
+
+	if plan, err := p.Solver.SolveHungarianOptimal(ctx, req); err != nil {
+		ctx.JSON(http.StatusInternalServerError, gin.H{"error": err})
+	} else {
+		ctx.JSON(http.StatusOK, gin.H{"plan": plan})
+	}
+}
+
 // Return top planning results to user
 func (p *MyPlanner) getPlanningApi(ctx *gin.Context) {
 	logger := iowrappers.Logger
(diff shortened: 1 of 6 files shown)


planner/planner.go
```diff
@@ -349,6 +349,20 @@ func (p *MyPlanner) homePageHandler(ctx *gin.Context) {
...
+	req := &PlanningReq{}
+
+	if err := ctx.ShouldBindJSON(req); err != nil {
+		ctx.JSON(http.StatusBadRequest, gin.H{"error": err})
+	}
+
+	if plan, err := p.Solver.SolveHungarianOptimal(ctx, req); err != nil {
+		ctx.JSON(http.StatusInternalServerError, gin.H{"error": err})
+	} else {
+		ctx.JSON(http.StatusOK, gin.H{"plan": plan})
```
rwangsc18, line 362: Are we just returning the new plan as text on html page for the moment?

planner/solver.go
```diff
@@ -282,7 +309,84 @@ func reversePlans(plans []PlanningSolution) []PlanningSolution {
...
+	if len(weights) == 0 {
+		return nil
+	}
+	fixedWeights := make([][]int, len(weights))
+	for idx := range fixedWeights {
+		fixedWeights[idx] = make([]int, len(weights[0]))
+	}
+	// assume the values are non-negative
+	var maxValue int
+	for _, ws := range weights {
```
rwangsc18, line 340: nit: if Hungarian algorithm supports negative value in the weight matrix, we can skip finding the maxValue here and use `-val` directly?

rwangsc18 (approved):
LGTM

planner/solver.go
timwangmusic: It appears that using negated values do not equivalent to solving a maximization problem.

planner/planner.go
timwangmusic: I am making a new commit so that we return more info including time slot, place name and category.

PR #301, opened 2023-08-20
Fix: discard erroneous city retrieval results from Redis

timwangmusic:
## Description
When Redis clients fail to retrieve city details, the entry in the slice is empty and the client has to filter it out.

## Solution
* Only pushes valid city details to the result slice.
* Workers send results to a channel for the main Go routine to collect.

## Testing
- [x] Added new unit tests

## Checks

### DIFF
--- a/test/redis_client_mocks/nearby_cities_search_test.go
+++ b/test/redis_client_mocks/nearby_cities_search_test.go
@@ -4,17 +4,17 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/weihesdlegend/Vacation-planner/iowrappers"
 	"reflect"
+	"sort"
 	"testing"
 )
 
 var expectedCities = []iowrappers.City{
 	{
-		ID:         "5376803",
-		GeonameID:  5376803,
-		Name:       "Newark",
-		Latitude:   37.52966,
-		Longitude:  -122.0402,
-		Population: 45000,
+		ID:         "5267812",
+		GeonameID:  5267812,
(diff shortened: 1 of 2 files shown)


test/redis_client_mocks/nearby_cities_search_test.go
```diff
@@ -58,6 +59,10 @@ func TestNearbyCitiesSearch_shouldReturnNearbyCities(t *testing.T) {
 		t.Fatalf("expected number of city returned equals %d, got %d", len(expectedCities), len(resultCities))
 	}
 
+	sort.Slice(resultCities, func(i, j int) bool {
```
rwangsc18, line 62: Does this unit test cover erroneous city retrieval scenario?

timwangmusic: The tests do not cover the failure scenario. Currently with mini-redis we are doing integration test, so with the interface it is hard to test when we fail to fetch from the database or deserialization.

rwangsc18: Got it, this is fine then. Do you know if redis provide some mock interface?

rwangsc18:
LGTM

rwangsc18 (approved):
LGTM

test/redis_client_mocks/nearby_cities_search_test.go
timwangmusic: It does not, that's the point they are making no mocks.

PR #307, opened 2023-09-05
Feat: Search with nearby cities

timwangmusic:
## Description
Implement the search with nearby cities feature.

## Solution
* Concurrently search for travel plans in nearby cities of the primary city.
* Google Maps API call reduction measures:
- [x]  Limit number of nearby cities within a certain threshold.
- [x]  Prioritize results from the primary city. Only include results from nearby cities when results from the primary city is not sufficient.
- [x]   Fall back to primary city results if nearby cities results have error.

## Testing
- [x] Added new unit tests

## Checks


planner/planner.go
```diff
@@ -280,8 +283,52 @@ func (p *MyPlanner) cityStatsHandler(context *gin.Context) {
 	context.JSON(http.StatusOK, view)
 }
 
-func (p *MyPlanner) Planning(ctx context.Context, planningRequest *PlanningReq, user string) (resp PlanningResponse) {
-	planningResponse := p.Solver.Solve(ctx, planningRequest)
+func (p *MyPlanner) Planning(ctx context.Context, planningRequest PlanningReq, user string, withNearbyCities bool) (resp PlanningResponse) {
+	logger := iowrappers.Logger
+	var planningResponse *PlanningResp
```
rwangsc18, line 288: This looks fine to me. 
For future improvement, we can also consider adding a common interface with a Solve() method. We can wrap the new multi-city solver logic into struct receiver. In Planning(), we can instantiate a solver instance based on the user input.

planner/planner.go
```diff
@@ -280,8 +283,52 @@ func (p *MyPlanner) cityStatsHandler(context *gin.Context) {
...
+					AdminAreaLevelOne: planningRequest.Location.AdminAreaLevelOne,
+					Country:           planningRequest.Location.Country,
+				},
+				Radius: float64(planningRequest.SearchRadius / 1000),
+				Filter: go_geonames.CityWithPopulationGreaterThan15000,
+			})
+		if err != nil {
+			return PlanningResponse{Err: err}
+		}
+		locations := MapSlice[iowrappers.City, POI.Location](nearbyCityResponse.Cities, toLocation)
```
rwangsc18, line 317: How many nearby cities do we usually get? To many nearby cities, can increase the load of the server if the traffic gets big. We can consider adding a cap to the number of parallel search requests.

rwangsc18 (approved):
Great work! The results also look good on the phone. Left a few comments that you may consider as future improvements.

planner/planner.go
timwangmusic: I'll add a boolean field `withNearbyCities` in the `PlanningReq` struct.

planner/planner.go
timwangmusic: Good point. Actually in design doc we proposed a limit. Any number in mind?

rwangsc18: We can go with 10 first, and set this cap as a configurable parameter and adjust later. Do you know if Heroku automatically increase the number of servers we are using if the computation load increases?

timwangmusic: I don't think Heroku does that. If we deploy on kubernetes it will do that.

planner/planner.go
```diff
@@ -280,67 +291,130 @@ func (p *MyPlanner) cityStatsHandler(context *gin.Context) {
...
-	if planningResponse.Err != nil {
-		resp.Err = planningResponse.Err
-		resp.StatusCode = planningResponse.ErrorCode
-		return
+	// prioritize planning results from the primary location
+	if resp.Err == nil && len(resp.TravelPlans) >= planningRequest.NumPlans {
+		return resp
+	}
+
+	var lat, lng float64
```
rwangsc18, line 308: nit: combine declaration and assignment into one line?

rwangsc18 (approved):
LGTM. Great work!

planner/planner.go
timwangmusic: Done

PR #308, opened 2023-09-06
Use exact price level in Google Maps search requests

timwangmusic:
## Description
We don't need to obtain places with price level higher than the requested. This reduces the number of places we need to save to the database when querying for lower price levels.

## Solution
* Simplify search radius adjustment code to use built-in functions.

## Testing
- [ ] Added new unit tests

## Checks


rwangsc18 (approved):
LGTM

PR #309, opened 2023-09-08
Allow one place to appear multiple times in the search result page

rwangsc18:
## Description
Improve the quality of plan recommendation page, so that the same location may appear in multiple suggested plan trip candidates. Currently, the same place is only allowed to appear once, and this may limit the number of suggested plans for certain cities.

For example, Q pop shop appears twice in the result plan list.
Screenshot 2023-09-09 at 3 24 55 PM

## Solution
* Add a counter for each place during the main search/solver algorithm
* The counter can be implemented by a hashMap 
* In future PR, we can add this as a configurable param with some refactor on solver.go  

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/planner/solver.go
+++ b/planner/solver.go
@@ -18,6 +18,7 @@ import (
 )
 
 const (
+	SamePlaceDupCountLimit                = 2
 	NumPlansDefault                       = 5
 	TopSolutionsCountDefault              = 5
 	DefaultPlaceSearchRadius              = 10000
@@ -262,8 +263,8 @@ func FindBestPlanningSolutions(ctx context.Context, placeClusters [][]matching.P
 		topSolutionsCount = TopSolutionsCountDefault
 	}
 
-  priorityQueue := &MinPriorityQueue[Vertex]{}
-  includedPlaces := make(map[string]bool)
+	priorityQueue := &MinPriorityQueue[Vertex]{}
+	includedPlaces := make(map[string]int8)
 	resp = make(chan PlanningResp, 1)
 
 	for iterator.HasNext() {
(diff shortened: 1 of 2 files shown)


planner/solver.go
```diff
@@ -487,27 +488,28 @@ func saveSolutions(ctx context.Context, c *iowrappers.RedisClient, req *Planning
...
 
-// removePlaces deletes all places of the input plan potentially stored in the input hashMap.
+// removePlaces reduces counts of places of the input plan potentially stored in the input hashMap.
 // This function can allow the main algorithm to re-introduce new plans into top candidates.
-func removePlaces(seenPlaces map[string]bool, plan PlanningSolution) {
+func removePlaces(includedPlaces map[string]int8, plan PlanningSolution) {
 	for _, placeID := range plan.PlaceIDS {
-		if _, exists := seenPlaces[placeID]; !exists {
+		count, exists := includedPlaces[placeID]
+		if !exists || (count == 0) {
```
timwangmusic, line 510: Is the second case possible?

planner/solver.go
```diff
@@ -262,8 +263,8 @@ func FindBestPlanningSolutions(ctx context.Context, placeClusters [][]matching.P
 		topSolutionsCount = TopSolutionsCountDefault
 	}
 
-  priorityQueue := &MinPriorityQueue[Vertex]{}
-  includedPlaces := make(map[string]bool)
+	priorityQueue := &MinPriorityQueue[Vertex]{}
+	includedPlaces := make(map[string]int8)
```
timwangmusic, line 267: Can you please add in a comment the rationale for using int8 values?

timwangmusic (commented):
The number of reuses is currently not loaded from configurations. Can we add that part?

planner/solver.go
rwangsc18: Done, added in 2137604.

planner/solver.go
rwangsc18: are you referring to (count == 0)? if so, yes it could happen. If one calls its function `removePlaces()` more than needed, such case can happen.

rwangsc18:
> The number of reuses is currently not loaded from configurations. Can we add that part?

I had some difficulty to load it through yaml config file at my initial attempt, so this PR will just focus on changing the place duplication limit from 1 to 2. I also need to do some refactoring for a few functions in solver.go, so we can make a clean pass of the limit from config file. It's easier to do this in a separate PR. wdyt?

timwangmusic (approved):
LGTM, nice work!

PR #312, opened 2023-09-08
Update Place freshness logic

timwangmusic:
## Description
Previously, we determine if a new API call to the Maps is necessary when the number of results is lower than requested. However, with the new Maps query fixed on min/max price level, this will not work properly since higher priced restaurants are inevitably less than the requested most of the time. Therefore the new logic says as long as we have some data for the location with the requested category and price level, it is acceptable data from the database.

## Solution
* `MapsSearchTime` entries have `PriceLevel` info now
* Update place freshness if statement

## Testing
- [x] Added new unit tests

## Checks

### DIFF
--- a/iowrappers/poi_searcher.go
+++ b/iowrappers/poi_searcher.go
@@ -134,27 +134,31 @@ func (s *PoiSearcher) NearbySearch(context context.Context, request *PlaceSearch
 	}
 	location := request.Location
 
-	var cachedPlaces, places []POI.Place
-	var err error
-	cachedPlaces, err = s.redisClient.NearbySearch(context, request)
-	if err != nil {
-		Logger.Error(err)
+	var savedPlaces, places []POI.Place
+	var placesErr error
+	savedPlaces, placesErr = s.redisClient.NearbySearch(context, request)
+	if placesErr != nil {
+		Logger.Error(placesErr)
 	}
 
-	Logger.Debugf("[request_id: %s] number of results from redis is %d", context.Value(ContextRequestIdKey), len(cachedPlaces))
+	Logger.Debugf("[request_id: %s] number of results from redis is %d", context.Value(ContextRequestIdKey), len(savedPlaces))
 
 	// update last search time for the city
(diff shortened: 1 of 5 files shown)


iowrappers/poi_searcher.go
```diff
@@ -144,17 +144,21 @@ func (s *PoiSearcher) NearbySearch(context context.Context, request *PlaceSearch
...
 	// update last search time for the city
-	lastSearchTime, cacheMiss := s.redisClient.GetMapsLastSearchTime(context, location, request.PlaceCat)
+	lastSearchTime, lastSearchTimeMiss := s.redisClient.GetMapsLastSearchTime(context, location, request.PlaceCat, request.PriceLevel)
 
 	currentTime := time.Now()
-	// use place data from database if the location is known and the data is fresh, and we have sufficient data
-	if cacheMiss == nil && (currentTime.Sub(lastSearchTime) <= MinMapsResultRefreshDuration && uint(len(cachedPlaces)) >= request.MinNumResults) {
-		Logger.Infof("[request_id: %s] Using Redis to fulfill request. Place Type: %s", context.Value(ContextRequestIdKey), request.PlaceCat)
+	// use place data from the database if the data is fresh, and we have at least one known place satisfying the request
+	if currentTime.Sub(lastSearchTime) <= MinMapsResultRefreshDuration && lastSearchTimeMiss == nil && len(cachedPlaces) > 0 {
```
rwangsc18, line 151: Should we first check (lastSearchTimeMiss == nil) here, in case lastSearchTime returns some invalid result?

rwangsc18: nit: we can also turn IF condition into a utility function? say `IsCacheResultFresh()`. Then we can make this line shorter and even remove the long comment.

rwangsc18 (approved):
Overall LGTM. left one comment.

iowrappers/poi_searcher.go
timwangmusic: Good point of making the IF condition as a utility function. Will do. The reason I am moving the `lastSearchTimeMiss == nil` to the second place is freshness is the top priority regardless of cache miss or not.

PR #313, opened 2023-09-12
Refactor: spin a go routine to close channels before server shut down

timwangmusic:
## Description
Previously a `os.Signal` channel blocks the main go routine and the `sync.Wait` is basically useless.

## Solution
* Dispatch event workers in the main go routine.
* Spin a go routine when receiving shutdown signals from OS. Close the event channel when shutdown signal is received.
* After workers are done the main go routine can continue.
* Finally we gracefully close the server so that no more requests can be sent.

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/main.go
+++ b/main.go
@@ -85,7 +85,7 @@ func RunServer() {
 		log.Fatal(err)
 	}
 
-	log.Info("Server gracefully shut down")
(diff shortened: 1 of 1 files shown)


rwangsc18 (approved):
LGTM

PR #315, opened 2023-09-16
Fix: blank page when no solution is found

timwangmusic:
## Description
Recently we added a guardrail against plan generator iterator having to process place clusters with zero place. However the error returned by the place cluster generator was not handled properly, leading to not redirecting to the root path.

## Solution
* Redirect to `/v1` when internal error happen.

## Testing
- [ ] Added new unit tests

## Checks


rwangsc18 (approved):
LGTM

PR #319, opened 2023-09-27
Experiement: use a new formula for places price level zero

timwangmusic:
## Description
This change affects all the places with **Visit** category, since all their current price levels are zero. Previously the place rating to pricing ratios for Visit places are treated flatly, and this affect the result of score since the final score will only depend on the number of user ratings. With this change, the ratings for Visit places will be taken into account.

## Solution
* Modify the rating-to-pricing ratio to rating divided by the maximum score (5.0).

## Testing
- [x] Added new unit tests

## Checks

### DIFF
--- a/matching/score.go
+++ b/matching/score.go
@@ -10,8 +10,7 @@ import (
 )
 
 const (
-	AvgRating  = 3.0
-	AvgPricing = PriceDefaultMean
+	MaxPlaceRating = 5.0
 )
 
 // OLD plan scoring method, use maxDist as the normalisation factor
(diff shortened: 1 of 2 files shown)


matching/score.go
```diff
@@ -43,7 +42,7 @@ func Score(places []Place, distNorm int) float64 {
 func singlePlaceScore(place Place) float64 {
 	var ratingPricingRatio float64
 	if place.PlacePrice() == 0 {
-		ratingPricingRatio = AvgRating / AvgPricing // set to average single Place rating-price ratio
+		ratingPricingRatio = float64(place.Rating()) / MaxPlaceRating
```
rwangsc18, line 45: nit: we can rename `ratingPricingRatio` to `boostFactor`, since when the place price is zero, it is more like a normalized rating.

rwangsc18 (approved):
LGTM. Just left one comment

matching/score.go
timwangmusic: Done.

PR #322, opened 2023-09-30
Add Google photo URL validation

rwangsc18:
## Description
Some place may contain invalid photo URL, such as "www.google.com". see the example of "Birk's". 
image

## Solution
Add a new URL validation function, and pass it into URL extraction function.
The new validator checks if the URL starts with a desired prefix. "https://lh3.googleusercontent.com/places/"

## Testing
- [] Integration testing on Heroku staging
- [x] Added new unit tests

## Checks

### DIFF
--- a/iowrappers/photos_client.go
+++ b/iowrappers/photos_client.go
@@ -10,6 +10,8 @@ import (
 	"golang.org/x/net/html"
 )
 
+const ValidPrefix = "https://lh3.googleusercontent.com/places/"
+
 type PhotoHttpClient struct {
 	// connect to Google Photo API
 	client     *http.Client
@@ -24,7 +26,7 @@ var isHtmlAnchor = func(node *html.Node) bool {
 }
 
 var isValidPhotoUrl = func(url string) bool {
-	return strings.HasPrefix(url, "https://lh3.googleusercontent.com/places/")
+	return strings.HasPrefix(url, ValidPrefix)
 }
 
 // CreatePhotoHttpClient is a factory method for PhotoClient
@@ -59,9 +61,9 @@ func (photoClient *PhotoHttpClient) GetPhotoURL(photoRef string) PhotoURL {
 	photoURL, err = parseHTML(body, isHtmlAnchor, isValidPhotoUrl)
 	if err != nil {
 		Logger.Warn("Err Msg: ", err.Error())
+		return ""
 	}
 	return photoURL
-
 }
 
 func parseHTML(htmlBody []byte, judger func(*html.Node) bool, validator func(string) bool) (PhotoURL, error) {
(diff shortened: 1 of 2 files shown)


rwangsc18:
Great news! after rebasing on the latest master, Node.js CI also passed! Thanks @timwangmusic for adding it.

iowrappers/photos_client.go
```diff
@@ -23,6 +23,10 @@ var isHtmlAnchor = func(node *html.Node) bool {
 	return node.Type == html.ElementNode && node.Data == "a"
 }
 
+var isValidPhotoUrl = func(url string) bool {
```
timwangmusic, line 26: Nit: can this be `const` instead of `var`?

iowrappers/photos_client.go
```diff
@@ -23,6 +23,10 @@ var isHtmlAnchor = func(node *html.Node) bool {
...
+	return strings.HasPrefix(url, "https://lh3.googleusercontent.com/places/")
```
timwangmusic, line 27: Nit: put the url in a const

iowrappers/photos_client.go
```diff
@@ -52,15 +56,15 @@ func (photoClient *PhotoHttpClient) GetPhotoURL(photoRef string) PhotoURL {
 		return photoURL
 	}
 
-	photoURL, err = parseHTML(body, isHtmlAnchor, "href")
+	photoURL, err = parseHTML(body, isHtmlAnchor, isValidPhotoUrl)
 	if err != nil {
```
timwangmusic, line 60: what do we return if no URL is found?

timwangmusic, line 60: We are not actually handling the error.

rwangsc18: Done. Add a return of empty string when error occurs.

iowrappers/photos_client.go
rwangsc18: It will give a compilation error when changed to const. I am not sure how function object is treated in golang.

iowrappers/photos_client.go
rwangsc18: I tried, but it would give a compilation error when changed to const, see leave it as "var" now. I am not fully sure how function object is treated in golang.

rwangsc18:
Addressed all previous comments in [64bdb01](https://github.com/timwangmusic/Vacation-Planner/pull/322/commits/64bdb013e4d13649fff0d27cedf38637fedcb1c4)

iowrappers/photos_client.go
timwangmusic: So if we return an empty string, how would the caller handle it?

rwangsc18: The caller will just have an empty string for the photo url at the moment, and the place photo will still be missing. This PR mainly deals with the issue where Photo URL extraction returns `www.google.com`. 
In a separate PR, we can add the caller's handling of empty url. For example, we can use a default picture or icon.

timwangmusic (approved):

rwangsc18:
Thanks for the review!

PR #326, opened 2023-10-02
Set number of pages per planning results received

timwangmusic:
## Description
We used to have a fixed number of pages (5 pages for 10 results) for plan template result section. This is not ideal when the number of results is less than or more than 10 since some pages will show blanks.

## Solution
* Set pages per planning results received.

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/assets/js/plan_template.js
+++ b/assets/js/plan_template.js
@@ -213,16 +213,17 @@ function parseResponse(response) {
                     })
                 }
             )
-            // only shows the first 3 results
+
             $('#tables .table').slice(itemsPerPage).hide();
             $('#pagination').show();
             $('#pagination').pagination({
 
                 // Total number of items to be paginated
-                items: totalItemsCount,
+                items: plansCount,
 
                 // Items allowed on a single page
                 itemsOnPage: itemsPerPage,
+                pages: Math.ceil(plansCount / itemsPerPage),
                 onPageClick: function (pageIdx) {
                     const itemsOnPage = 2;
                     $('#tables .table').hide()


rwangsc18:
Um... I am a bit confused by the PR description. Do we already have multiple pages in the result html? I always think that we only show a single page for the top 5 results.

timwangmusic:
> Um... I am a bit confused by the PR description. Do we already have multiple pages in the result html? I always think that we only show a single page for the top 5 results.

This change is for the plan template page only.

rwangsc18:
> > Um... I am a bit confused by the PR description. Do we already have multiple pages in the result html? I always think that we only show a single page for the top 5 results.
> 
> This change is for the plan template page only.

Right, I was asking about the template page. For example, is it what you called a template page https://www.unwind.dev/v1/plans?date=2023-10-02&location=Los+Angeles%2C+CA%2C+USA&precise=false&nearby=false&price=2. Pages like this always show 5 results or less. But it seems based on your PR, there are a few pages like this.

timwangmusic:
> 
> 
> 
> 
> 
> 
> > > Um... I am a bit confused by the PR description. Do we already have multiple pages in the result html? I always think that we only show a single page for the top 5 results.
> 
> > 
> 
> > This change is for the plan template page only.
> 
> 
> 
> Right, I was asking about the template page. For example, is it what you called a template page https://www.unwind.dev/v1/plans?date=2023-10-02&location=Los+Angeles%2C+CA%2C+USA&precise=false&nearby=false&price=2. Pages like this always show 5 results or less. But it seems based on your PR, there are a few pages like this.

No, this is the template page: https://www.unwind.dev/v1/template

rwangsc18 (approved):
LGTM

PR #327, opened 2023-10-05
Enhancement: collect planning solutions when timeout happens

timwangmusic:
## Description
When users query for a large number of time slots (>4) a regular server typically fail to examine all the possible solutions. We could consider collect solutions when computational timeout happen.

## Solution
* Collect planning solutions when computational timeout happens.
* Move timeout context to `FindBestPlanningSolutions` method since its caller do not need to be aware of this mechanism.
* Remove computational timeout error.
* Increase front-end timeout from 10s to 15s.

## Testing
- [x] Added new unit tests

## Checks

### DIFF
--- a/assets/js/plan_template.js
+++ b/assets/js/plan_template.js
@@ -140,7 +140,7 @@ async function postPlanTemplate() {
     console.log(`data about to send: ${JSON.stringify(data)}`);
 
     axios.post(
-        url, JSON.stringify(data), { timeout: 10000 }
+        url, JSON.stringify(data), { timeout: 15000 }
     ).then(
         function (response) {
             console.log(response.data);
--- a/planner/solver.go
+++ b/planner/solver.go
@@ -24,7 +24,6 @@ const (
 	TopSolutionsCountDefault              = 5
 	DefaultPlaceSearchRadius              = 10000
 	CategorizedPlaceIterInitFailureErrMsg = "categorized places iterator init failure"
-	ComputationTimedOutErrMsg             = "computation timed out"
 	ErrMsgMismatchIterAndPlace            = "mismatch in iterator status vector length"
 	ErrMsgRepeatedPlaceInSameTrip         = "repeated places in the same trip"
 )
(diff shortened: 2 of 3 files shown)


planner/solver.go
```diff
@@ -333,21 +332,23 @@ func createPlanningSolutionCandidate(placeIndexes []int, placeClusters [][]match
...
+	resp = &PlanningResp{}
 
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, SolverTimeout)
+	defer cancel()
 	for iterator.HasNext() {
 		select {
-		case <-ctx.Done():
-			resp <- PlanningResp{Err: errors.New(ComputationTimedOutErrMsg), ErrorCode: RequestTimeOut}
-			return
+		case <-ctxWithTimeout.Done():
```
rwangsc18, line 349: Since we changed the response under timeout, is it easy to add a unit test to cover this scenario, i.e. ctxWithTimeout?

rwangsc18 (commented):
Overall look good to me. Just left one comment

planner/solver.go
timwangmusic: It is not easy to test, since the timeout happens inside the `FindBestPlanningSolutions` function. If you run locally you can see the error message when timeout happens.

sananand007 (approved):
lg

assets/js/plan_template.js
```diff
@@ -140,7 +140,7 @@ async function postPlanTemplate() {
     console.log(`data about to send: ${JSON.stringify(data)}`);
 
     axios.post(
-        url, JSON.stringify(data), { timeout: 10000 }
+        url, JSON.stringify(data), { timeout: 15000 }
```
sananand007, line 143: why increase ?

timwangmusic: Good question, this makes sure the front end requests do not timeout before backend computations time outs happen.

planner/solver.go
rwangsc18: Sg, lgtm

rwangsc18 (approved):

xtpitt (approved):
LGTM

PR #328, opened 2023-10-05
Add logo PNG

timwangmusic:
## Description
So we can get rid of the staled `pr_logo` branch.


sananand007 (approved):
lgtm. Once you merge and delete we should see the file in the default branch. that is the expectation

PR #331, opened 2023-10-17
[Enhancement] Sort places by score before solving combinatorics

timwangmusic:
## Description
Now we are doing early termination when solving combinatorics. So the idea is if the input place lists are sorted by score, the solver would be able to check the places with higher score first, resulting in higher quality plans in general.

## Solution
* export the function to compute single place score to the higher level packages
* sort places by score before solving the combination problem.

## Testing
- [x] Added new unit tests

## Checks

### DIFF
--- a/planner/solver.go
+++ b/planner/solver.go
@@ -1,6 +1,7 @@
 package planner
 
 import (
+	"cmp"
 	"container/heap"
 	"context"
 	"errors"
@@ -537,6 +538,8 @@ func (s *Solver) generatePlacesForSlots(ctx context.Context, req *PlanningReques
 		if len(placesByPrice) == 0 {
 			return nil, fmt.Errorf("failed to find any place for category %s at slot %s for location %+v", slot.Category, slot.TimeSlot.ToString(), req.Location)
 		}
+		// sort places by score descending so the solver checks places with higher score first
+		slices.SortFunc(placesByPrice, func(a, b matching.Place) int { return cmp.Compare(matching.PlaceScore(b), matching.PlaceScore(a)) })
 		placeClusters = append(placeClusters, placesByPrice)
 	}
 	return placeClusters, nil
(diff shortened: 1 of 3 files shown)


planner/solver.go
```diff
@@ -537,6 +538,8 @@ func (s *Solver) generatePlacesForSlots(ctx context.Context, req *PlanningReques
 		if len(placesByPrice) == 0 {
 			return nil, fmt.Errorf("failed to find any place for category %s at slot %s for location %+v", slot.Category, slot.TimeSlot.ToString(), req.Location)
 		}
+		// sort places by score descending so the solver checks places with higher score first
+		slices.SortFunc(placesByPrice, func(a, b matching.Place) int { return cmp.Compare(matching.PlaceScore(b), matching.PlaceScore(a)) })
```
rwangsc18, line 542: Is it simple to add some test to check that places are sorted in slices in the solver?

rwangsc18 (commented):
Great idea! This is indeed should be a good change. Just Left one comment in the PR. Plx take a look.

planner/solver.go
timwangmusic: We could use [github.com/uber-go/mock](https://github.com/uber-go/mock) to mock `iowrappers.SearchClient` and the `Matcher` interface, it won't be easy though. Let me know if you think we should try it in this PR or in another follow-up.

timwangmusic: By looking closer at the specific issue, I found that a deeper problem is there should be a "super" place matcher that employ the [strategy pattern](https://refactoring.guru/design-patterns/strategy/go/example), so the solver can just call match once in the super matcher. I'll create a PR to address this.

rwangsc18: Sg, no need to over-complicate the change in one PR. My original thought was that maybe we can just add a test to verify results from findOptimalSolution are sorted.

rwangsc18: I forgot to add that it is okay to add this test in a separate PR.

rwangsc18 (approved):
LGTM

PR #335, opened 2023-10-29
Make all default return of photo URLs to be empty string

rwangsc18:
## Description
Make all default return of photo URLs to be empty string

## Solution
* Align all return values to be an empty string when failed processing happens 

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/iowrappers/photos_client.go
+++ b/iowrappers/photos_client.go
@@ -46,16 +46,17 @@ func (photoClient *PhotoHttpClient) GetPhotoURL(photoRef string) PhotoURL {
 	res, err := photoClient.client.Get(reqURL)
 	if err != nil {
 		Logger.Fatal(err)
-		return photoURL
+		return ""
 	}
 	body, err := io.ReadAll(res.Body)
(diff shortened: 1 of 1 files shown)


rwangsc18:
I also looked into how to serve a default icon/picture. Do we have example like that in this codebase?

timwangmusic:
> I also looked into how to serve a default icon/picture. Do we have example like that in this codebase? 

We are serving icons in the planning results right?

timwangmusic (changes requested):
If we return empty strings, how would the front-end render them? I think a better mitigation is to use a default picture when we cannot get the photos. In that case, this function should return an error as the last returning value.

rwangsc18:
> If we return empty strings, how would the front-end render them? I think a better mitigation is to use a default picture when we cannot get the photos. In that case, this function should return an error as the last returning value.

I was thinking along the same line. I have found some example like this for HTML to use a local image (checked in repo), but it does not work right away. I also searched the codebase and can't find a similar example.

PR #338, opened 2023-11-05
Add implementation for GetPhotoURL for MapsPhotoClient

rwangsc18:
## Description
Implement for GetPhotoURL for MapsPhotoClient, use [maps client function call](https://pkg.go.dev/googlemaps.github.io/maps@v1.5.0#Client.PlacePhoto). 

## Solution
* Modify trip plan detail HTML template, and accept image encoded base64 format.
* Update PhotoClient interface to accept ctx, because maps client call require a context variable

## Testing
- [ ] Added new unit tests

## Checks


rwangsc18:
Tested locally, it works well. I also pushed to staging env https://staging-vp.herokuapp.com/v1/, but I had some difficulty login in staging env (forget my login credentials again..)

iowrappers/photos_client.go
```diff
@@ -23,13 +28,14 @@ var isValidPhotoUrl = func(url string) bool {
 type PhotoURL string
 
 type PhotoClient interface {
-	GetPhotoURL(string) PhotoURL
+	GetPhotoURL(context.Context, string) PhotoURL
```
timwangmusic, line 31: Suggest return `(PhotoURL, error)`

iowrappers/photos_client.go
```diff
@@ -41,7 +47,11 @@ type PhotoHttpClient struct {
 // CreatePhotoClient is a factory method for PhotoClient
 func CreatePhotoClient(apiKey string, baseURL string, enableMapPhotoClient bool) PhotoClient {
```
timwangmusic, line 48: Suggest return `(PhotoClient, error)`, otherwise the caller would not know if the client is valid.

planner/planner.go
```diff
@@ -618,17 +617,17 @@ func (p *MyPlanner) getPlanDetails(ctx *gin.Context) {
...
 
-func (p *MyPlanner) asyncGetTripRespPlaceDetails(wg *sync.WaitGroup, resp *PlaceDetailsResp, place POI.Place) {
-	*resp = p.getTripFromPlace(place)
+func (p *MyPlanner) asyncGetTripRespPlaceDetails(ctx context.Context, wg *sync.WaitGroup, resp *PlaceDetailsResp, place POI.Place) {
+	*resp = p.getTripFromPlace(ctx, place)
 	wg.Done()
 }
 
-func (p *MyPlanner) getTripFromPlace(place POI.Place) PlaceDetailsResp {
+func (p *MyPlanner) getTripFromPlace(ctx context.Context, place POI.Place) PlaceDetailsResp {
```
timwangmusic, line 625: The function needs to return an error as well.

iowrappers/photos_client.go
rwangsc18: I think it might be okay to return PhotoURL. Since most of errors come from the place does not have valid image data, we can handle the exception by returning image of a default image.

rwangsc18: Previously I had trouble sending a default photo, because we don't have a file server to host such as default picture. Now it should be easier, since we encode the image to some binary string and embed them html response. I will add this default photo logic in a following PR.

iowrappers/photos_client.go
rwangsc18: Sg, will do.

iowrappers/photos_client.go
rwangsc18: Feel free to let me know what you think. We can also handle this error on the caller side.

planner/planner.go
rwangsc18: This comment depends on whether we decide on the function signature `GetPhotoURL()`.

iowrappers/photos_client.go
timwangmusic: > Previously I had trouble sending a default photo, because we don't have a file server to host such as default picture. Now it should be easier, since we encode the image to some binary string and embed them html response. I will add this default photo logic in a following PR.

We do have a file server to host images, in `/assets` directory

rwangsc18: Somehow it does not work simply by adding a local file to /assets folder. We will need to think about some other option to load a default picture.

rwangsc18: We eventually want to catch and handle the exception somewhere. I think the current implementation (returning PhotoURL instead of (PhotoURL, error)) is okay, because we need to return something in case Google maps API fails to provide valid data. The exception handling logic is done inside GetPhotoURL().

> Right, since the function `GetPhotoURL` is making an external HTTP call, it should return an error. The caller of this function should handle that error.

iowrappers/photos_client.go
rwangsc18: Done, see [839254e]

planner/planner.go
rwangsc18: If we decide that GetPhotoURL() from PhotoClient handles error cases internally and always return some valid photo data, we probably don't need to return an error in this function.

iowrappers/photos_client.go
timwangmusic: The exception handling should not be done in this function, the caller should check the error and use a default URL if we do not have mitigation yet.

planner/planner.go
timwangmusic: This function should handle the error and provide the response with a default URL if needed

rwangsc18: Done, see [4d92fba](https://github.com/timwangmusic/Vacation-Planner/pull/338/commits/4d92fbacbfcac0d0847f62a05acaa06e0f8143d6)

iowrappers/photos_client.go
rwangsc18: Done, see [4d92fba](https://github.com/timwangmusic/Vacation-Planner/pull/338/commits/4d92fbacbfcac0d0847f62a05acaa06e0f8143d6)

iowrappers/photos_client.go
```diff
@@ -23,13 +28,14 @@ var isValidPhotoUrl = func(url string) bool {
...
-	GetPhotoURL(string) PhotoURL
+	GetPhotoURL(context.Context, string) (PhotoURL, error)
 }
 
 // Use Google Map API
 type MapsPhotoClient struct {
-	apiKey     string
-	apiBaseURL string
+	maps_client *maps.Client
+	apiKey      string
```
timwangmusic, line 37: Nit: Now we are using `maps.Client`, maybe `apiKey` is not required anymore?

timwangmusic (approved):
Awesome work! LGTM

iowrappers/photos_client.go
```diff
@@ -117,7 +124,27 @@ func dfs(node *html.Node, judger func(*html.Node) bool, validator func(string) b
 	return "", false
 }
 
-// TODO(rwangsc18): add real implementation
-func (photoClient *MapsPhotoClient) GetPhotoURL(photoRef string) PhotoURL {
-	return PhotoURL(photoRef + "fake_url")
+func (photoClient *MapsPhotoClient) GetPhotoURL(ctx context.Context, photoRef string) (PhotoURL, error) {
+	// get PlacePhotoResponse
```
timwangmusic, line 128: Nit: remove this comment.

rwangsc18: Done, removed in 60ad77a

rwangsc18:
spot check some search results in deployed stage vp

PR #340, opened 2023-11-11
Fix: trip details carousel button getting stuck

timwangmusic:
## Description
The plan details page carousel did not consider when we have multiple places in a travel plan. There were 3 fixed button in the HTML page. After users click on the last image, the screen will stuck.

## Solution
* Render number of carousel buttons per places in the travel plan
* Use dark variant for the carousel indicators
* Reformat the HTML page source code
* Reformat conditional class assignment such as `<div class={{if index $active $i}} "card-body" {{else}} "card-body d-none" {{end}}>` so that we can do the reformat

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/assets/templates/trip_plan_details_template.html
+++ b/assets/templates/trip_plan_details_template.html
@@ -1,126 +1,187 @@
 <!DOCTYPE html>
 <html lang="en">
-
-<head>
-    <meta charset="UTF-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <!--Bootstrap CSS-->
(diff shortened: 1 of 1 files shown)


rwangsc18 (commented):
Overall LGTM. Thanks for the fix to make carousel to show more than 3 photos!

Since photos from different places may have different height, I am not sure if the fade works well for them. It may create illusion of large latency when transitioning from places.

timwangmusic:
> Overall LGTM. Thanks for the fix to make carousel to show more than 3 photos!
> 
> 
> 
> Since photos from different places may have different height, I am not sure if the fade works well for them. It may create illusion of large latency when transitioning from places. 

Agreed, should we fix the image height or remove the fading effect?

rwangsc18:
> Agreed, should we fix the image height or remove the fading effect?
I would recommend removing the fading effect. Images come in with different height, aligning them with the height will stretch some to a weird height:width ratio.

timwangmusic:
> > Agreed, should we fix the image height or remove the fading effect?
> > I would recommend removing the fading effect. Images come in with different height, aligning them with the height will stretch some to a weird height:width ratio.

Removed the fading effect.

rwangsc18 (approved):
LGTM

PR #343, opened 2023-11-16
Fix: get plan details handler function

timwangmusic:
## Description
There is a potential issue that the previous location's place details from one location "leaks" to the next location's place details. For example, the place editorial summary is copied to the next location as users move across the slides.

## Solution
* The most important fix in this PR is create a new variable for each retrieved place instead of reusing the same one.
```		
var place POI.Place
cacheErr = p.RedisClient.FetchSingleRecord(ctx, placeKey, &place)

go p.asyncGetTripRespPlaceDetails(&wg, &tripResp.PlaceDetails[idx], place)
```

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/planner/planner.go
+++ b/planner/planner.go
@@ -558,64 +558,61 @@ func (p *MyPlanner) getPlanningApi(ctx *gin.Context) {
 }
 
 func (p *MyPlanner) getPlanDetails(ctx *gin.Context) {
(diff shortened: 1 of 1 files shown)


xtpitt (approved):
LGTM

PR #347, opened 2023-12-04
Remove username parameter in user profile requests

timwangmusic:
## Description
Having username does not achieve anything since the backend relies on cookies to identify users. So it is cleaner and less confusing to users for not including the username parameter in user profile requests.

## Solution
* The other changes are formatting with Prettier.

## Testing


rwangsc18 (approved):
LGTM

PR #350, opened 2023-12-17
Experiment: Update default search radius to 20km

timwangmusic:
## Description
Previously the default search radius is 10km, which is quite limited for larger cities. The average radius of the 10 most populated cities in the US is around 18.84 miles.

## Solution
* Increase default search radius to 20km (~ 12.34 miles) so we consider more places for planning requests.
* Address a TODO item where the computation for best planning results should consider search radius from input instead of using the default.

## Testing
- [ ] Added new unit tests

## Checks


sananand007 (approved):
lgtm, Is this not going to increase time to search and cache size etc ??

timwangmusic:
> lgtm, Is this not going to increase time to search and cache size etc ?? 

This only tunes the nearby search radius.

PR #352, opened 2023-12-20
Refactor: move home page art in its own file

timwangmusic:
## Description
Move home page art SVG image in its own file

## Solution
* Create a new file in `assets/public` and use it in the search page HTML file.
* Use prettier to reformat the search page HTML file.

## Testing


sananand007 (approved):
lg

timwangmusic:
This change breaks the view on mobile, closing it till we find a better solution.

PR #353, opened 2024-01-02
Fix: adding the first row in plan template does not give hour options

timwangmusic:
## Description
When removing all the rows in plan templates and adding the first row, the options in start and end time is blank.

## Solution
* Set default starting hour to be 9AM. 

## Testing

### DIFF
--- a/assets/js/plan_template.js
+++ b/assets/js/plan_template.js
@@ -6,6 +6,7 @@ import { logOut, updateUsername } from "./user.js";
 const username = updateUsername();
 const totalItemsCount = 10;
 const itemsPerPage = 2;
+const defaultStartingTime = "9";
 
 setDateToday();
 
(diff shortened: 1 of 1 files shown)


assets/js/plan_template.js
```diff
@@ -36,7 +37,10 @@ function insertNewRow(start = "8", end = "10", category = "Visit") {
   let cafeOption = document.createElement("option");
   let attractionOption = document.createElement("option");
 
-  if (template.rows.length > 2) {
+  // the number of rows will never be 1 or less
+  if (template.rows.length == 2) {
+    start = defaultStartingTime;
+  } else if (template.rows.length > 2) {
```
sananand007, line 43: why not just ? do we need a `if (condition)` 
```
else { }
```

sananand007 (approved):
lg, just a small comment.

PR #355, opened 2024-01-16
Fix: typos and reformat about page

timwangmusic:
## Description
Fix typos and reformat about page with Prettier.


sananand007 (approved):
lg

PR #357, opened 2024-01-29
Improve saving and retrieving planning solutions

timwangmusic:
## Description
Made saving and retrieving planning solutions more robust and less error prone.

## Solution
* When saving planning solutions, we should first make sure the plan is saved before saving its ID. We are now also cleaning up previous plan IDs if they exist.
* When retrieving planning solutions, we should make sure the plan details are fetched before adding them to the results. Previously we could end up with incomplete results.

## Testing
- [x] Added new unit tests

## Checks

### DIFF
--- a/iowrappers/redis_client.go
+++ b/iowrappers/redis_client.go
@@ -79,7 +79,7 @@ func (r *RedisClient) CollectPlanningAPIStats(event PlanningEvent) {
 	pipeline.PFAdd(RedisClientDefaultBlankContext, NumVisitorsPlanningAPI, event.User)
 
 	// set expiration time
-	if _, err := pipeline.Exists(RedisClientDefaultBlankContext, NumVisitorsPlanningAPI).Result(); err != nil && err == redis.Nil {
+	if _, err := pipeline.Exists(RedisClientDefaultBlankContext, NumVisitorsPlanningAPI).Result(); err != nil && errors.Is(err, redis.Nil) {
 		pipeline.Expire(RedisClientDefaultBlankContext, NumVisitorsPlanningAPI, PlanningStatExpirationTime)
 	}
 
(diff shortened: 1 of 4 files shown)


iowrappers/redis_client.go
```diff
@@ -611,32 +611,42 @@ func (r *RedisClient) SavePlanningSolutions(context context.Context, request *Pl
...
+	}
+
 	var recordKeys []string
 	for _, record := range request.PlanningSolutionRecords {
 		solutionRedisKey := strings.Join([]string{TravelPlanRedisCacheKeyPrefix, record.ID}, ":")
 		json_, err := json.Marshal(record)
 		if err != nil {
-			return err
+			Logger.Error(err)
+			continue
```
sananand007, line 628: why `continue` everywhere ? Is it to make sure we go through all the `redisKeys` ?

sananand007 (approved):
lg, just have some questions

iowrappers/redis_client.go
```diff
@@ -611,32 +611,42 @@ func (r *RedisClient) SavePlanningSolutions(context context.Context, request *Pl
...
 	}
 
+	// cleans up previous results
+	exists, _ := r.client.Exists(ctx, redisListKey).Result()
+	if exists == 1 {
+		if err := r.client.Del(ctx, redisListKey).Err(); err != nil {
+			Logger.Error(err)
+		}
+	}
+
```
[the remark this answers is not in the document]
timwangmusic: We are deleting `redisListKey`, which stores the plan IDs. The reason is that the LPush operation will keep appending to this key if it exists.

iowrappers/redis_client.go
timwangmusic: Correct. We don't want return an error and miss the rest of the plans like we did before.

PR #358, opened 2024-02-02
Fix: avoid re-generating plans when results are less than requested

timwangmusic:
## Description
When users select higher price levels, the number of candidate places become less. In these cases, the number of available plans are sometimes less than requested. As the same set of candidate places that are used in generating the plans, we are repeatedly generating the same plans over and over again. This is not scalable and wasting computation/storage resources.

## Solution
* Remove the constraint that the number of saved plans has to be larger than requested.
* `RedisClient.PlanningSolutions` method now returns a pointer instead of a struct.
* Add request ID when users log in.

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/iowrappers/redis_client.go
+++ b/iowrappers/redis_client.go
@@ -646,9 +646,9 @@ func (r *RedisClient) SavePlanningSolutions(ctx context.Context, request *Planni
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


sananand007 (approved):
lg

Issue #359, opened 2024-02-06
Remove orphan plans in Redis


timwangmusic:
When a new set of travel plans are generated, they can be found only through a parent Redis key, which expires in 24 hours (by current config). After the parent key expires, the plans cannot be referenced easily unless through saved plans list in the user profile. When users delete the references, they are likely to accumulate in the database. This is a flaw of the original design.

timwangmusic:
We still need a simple script or temporary endpoint to clean up existing orphan plans.

PR #360, opened 2024-02-06
Use user saved plan to generate plan details page

timwangmusic:
## Description
Related [issue 359](https://github.com/timwangmusic/Vacation-Planner/issues/359). Note that this is a **breaking** change, since users will not be able to check plan details for the previously saved plans.

## Solution
* Enhance user saved plans to generate the plan details page.
* Set expiration time on system generated plans.

## Verifications
* Save a plan from the result page -> go to details -> save the same plan from there -> verify in the profile that there should be no duplicate
* Save a plan directly from the details page -> go to the result page >  save the same plan from there -> verify in the profile that there should be no duplicate

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/iowrappers/redis_client.go
+++ b/iowrappers/redis_client.go
@@ -627,7 +627,7 @@ func (r *RedisClient) SavePlanningSolutions(ctx context.Context, request *Planni
 			Logger.Error(err)
 			continue
 		}
-		_, recordSaveErr := r.client.Set(ctx, solutionRedisKey, json_, 0).Result()
+		_, recordSaveErr := r.client.Set(ctx, solutionRedisKey, json_, PlanningSolutionsExpirationTime).Result()
 		if recordSaveErr != nil {
 			Logger.Error(err)
 			continue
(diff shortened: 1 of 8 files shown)


iowrappers/redis_client.go
```diff
@@ -627,7 +627,7 @@ func (r *RedisClient) SavePlanningSolutions(ctx context.Context, request *Planni
 			Logger.Error(err)
 			continue
 		}
-		_, recordSaveErr := r.client.Set(ctx, solutionRedisKey, json_, 0).Result()
+		_, recordSaveErr := r.client.Set(ctx, solutionRedisKey, json_, PlanningSolutionsExpirationTime).Result()
```
sananand007, line 630: This was defined on the previous PR I believe

sananand007 (approved):
lg, did you deploy it, Can I check the test deployment ?

timwangmusic:
> lg, did you deploy it, Can I check the test deployment ? 

I'll deploy to testing, however I think we need to make some tweaks to this change to make the maps working

iowrappers/redis_client.go
timwangmusic: No, the constant `PlanningSolutionsExpirationTime` existed before if this is what you mean.

PR #363, opened 2024-02-09
Add title to markers in Maps

timwangmusic:
## Description
Add place names as Map marker titles so users can see them while hanging over the markers.

## Solution
* Set `title` option when constructing markers.

## Testing


sananand007 (approved):
lg

PR #365, opened 2024-02-20
Configure production logger level to debug and above

timwangmusic:
## Description
Configure production logger level to debug and above since we need more logging in production.

## Solution
* Use `zap.NewProductionConfig()` and config the logging level instead of `zap.NewProduction()` which limits the logging level at `INFO`.

## Testing


sananand007 (approved):
lg

PR #371, opened 2024-04-18
Fix: race condition for AddCities method

timwangmusic:
## Description
The Go test tool detected that having concurrent access to the known_cities_name_to_id redis map results in race condition. Instead of using transactions which will not provide better performance than sequential executions, this change removes concurrency pattern all together for this function.

## Solution
* Remove concurrency pattern from the `AddCities` method.

## Testing
- `Go test --race` shows that the race condition is gone after the change.

## Checks


sananand007 (approved):
lg

PR #373, opened 2024-04-24
Async jobs for generating planning solutions

timwangmusic:
## Description
Previously when a user requests plans that are not generated and stored in the database, the application server computes new plans on the fly. Users will have to wait around 10s to get the results. This is ok for initial queries. When users change any single parameter, such as the price level, they will have to wait since those results are not available even though the only difference is the price level.

In this change, we launch planning solution workers to handle offline jobs. As a start, we enable computing solutions for all the price levels.

## Solution
* Created a worker dispatcher that runs when the server starts running.
* Created planning solutions workers that handle the async jobs.
* Use Golang's `RWMutex` to guard against concurrent job creations and executions.

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/iowrappers/jobs.go
+++ b/iowrappers/jobs.go
@@ -1,0 +1,74 @@
+package iowrappers
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+)
+
+type JobStatus string
+
+const (
+	JobStatusCreated   JobStatus = "created"
+	JobStatusRunning   JobStatus = "running"
+	JobStatusFailed    JobStatus = "failed"
+	JobStatusCompleted JobStatus = "completed"
+	JobExpirationTime            = time.Hour
+)
+
+type Job struct {
+	ID          string      `json:"id"`
+	Name        string      `json:"name"`
+	Description string      `json:"description"`
+	Parameters  interface{} `json:"parameters"`
+	Status      JobStatus   `json:"status"`
+	CreatedAt   time.Time   `json:"created_at"`
+	UpdatedAt   time.Time   `json:"updated_at"`
+}
+
+const JobRedisKeyPrefix = "job:"
+
+func (r *RedisClient) UpdateJob(ctx context.Context, job *Job) error {
+	key := JobRedisKeyPrefix + job.ID
+	curTime := time.Now()
+	if exists, err := r.Get().Exists(ctx, key).Result(); err != nil {
+		return err
+	} else if exists == 0 {
+		job.CreatedAt = curTime
+	} else {
+		job.UpdatedAt = curTime
+	}
+
+	data, err := json.Marshal(job)
+	if err != nil {
+		return err
+	}
+
+	return r.Get().Set(ctx, key, string(data), JobExpirationTime).Err()
+}
+
+func (r *RedisClient) GetJob(ctx context.Context, id string) (*Job, error) {
+	key := JobRedisKeyPrefix + id
+	if exists, err := r.Get().Exists(ctx, key).Result(); err != nil {
+		return nil, err
+	} else if exists == 0 {
+		return nil, fmt.Errorf("job %s does not exist", id)
+	}
+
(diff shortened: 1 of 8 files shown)


planner/dispatcher.go
```diff
@@ -0,0 +1,51 @@
+package planner
+
+import (
+	"context"
+	"github.com/weihesdlegend/Vacation-planner/iowrappers"
+	"sync"
+)
+
+const NumWorkers = 10
+
```
sananand007, line 10: why 10, ? and where do the workers run ?

planner/workers.go
```diff
@@ -0,0 +1,113 @@
+package planner
+
```
sananand007, line 2: so the workers are going to receive the Job and run it

sananand007 (approved):
All good. 

Can we have a quick meeting today to explain the PR please

planner/planner.go
```diff
@@ -498,6 +506,25 @@ func (p *MyPlanner) SetPlanSavedStatusForUser(ctx *gin.Context, numResults int,
 	return nil
 }
 
+func (p *MyPlanner) queuePlanningJobsForAllPriceLevels(req *PlanningRequest) error {
+	curTime := time.Now()
+	for l := range []POI.PriceLevel{POI.PriceLevelZero, POI.PriceLevelOne, POI.PriceLevelTwo, POI.PriceLevelThree, POI.PriceLevelFour} {
+		newReq := *req
+		newReq.PriceLevel = POI.PriceLevel(l)
+
```
sananand007, line 514: what is this portion of the code doing ? Lets discuss over the meeting

planner/dispatcher.go
timwangmusic: 10 is just a tentative number. We can increase this or put it in configurations if needed.

planner/workers.go
timwangmusic: yes, they are running in the background and ready for tasks from the job queue.

planner/dispatcher.go
sananand007: nice.. is this related to scalability in the sense , tune based on how fast or slow the results are returned back based on the number of `workers` we have.

planner/workers.go
sananand007: and each worker session is separate based on the user . So let's say we have `10` users logged in at the same time, and they all are triggering jobs we might have `100` jobs running on the backend which might be very rare right now, but trying to visualize ?

iowrappers/jobs.go
```diff
@@ -0,0 +1,83 @@
...
+type JobStatus string
+
+const (
+	JobStatusNew       JobStatus = "new"
+	JobStatusCreated   JobStatus = "created"
+	JobStatusRunning   JobStatus = "running"
+	JobStatusFailed    JobStatus = "failed"
+	JobStatusCompleted JobStatus = "completed"
+
+	JobExpirationTime = time.Hour
```
sananand007, line 19: so this is `1` hour ? any reason, or this can also be fine tuned ?

planner/dispatcher.go
```diff
@@ -0,0 +1,51 @@
...
+	return &Dispatcher{
+		JobQueue: make(chan *iowrappers.Job),
+		workers:  make([]*PlanningSolutionsWorker, 0),
+		solver:   s,
+		wg:       &sync.WaitGroup{},
+	}
+}
+
+func (d *Dispatcher) Run(ctx context.Context) {
+	mu := &sync.RWMutex{}
```
sananand007, line 28: This is nice, so basically regardless of how many users create new Jobs, they are always locked !

planner/planner.go
timwangmusic: It queues jobs for all price levels given other planning specs intact

iowrappers/jobs.go
timwangmusic: changed to 24 hours, so that admins can inspect.

PR #377, opened 2024-05-25
Set up router rate limiter

timwangmusic:
## Description
Set up router rate limiter with a middleware.

## Solution
* Limit requests per IP per hour to 100. We can make this configurable later.

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/planner/planner.go
+++ b/planner/planner.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"github.com/google/uuid"
 	"github.com/modern-go/reflect2"
+	"github.com/ulule/limiter/v3"
+	sredis "github.com/ulule/limiter/v3/drivers/store/redis"
 	"golang.org/x/oauth2"
 	"html/template"
 	"io"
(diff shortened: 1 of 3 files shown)


planner/planner.go
```diff
@@ -1014,6 +1018,22 @@ func (p *MyPlanner) GetPlaceDetails(ctx *gin.Context) {
 	ctx.JSON(http.StatusOK, place)
 }
 
+func (p *MyPlanner) rateLimiter() gin.HandlerFunc {
+	logger := iowrappers.Logger
+	// 100 requests per hour
+	rate, err := limiter.NewRateFromFormatted("100-H")
```
sananand007, line 1024: 100 per hours. ?

sananand007 (approved):
lg

planner/planner.go
sananand007: Can we test this now ?

timwangmusic: I tested for lower limit. It works pretty well. Currently it is set to 100 per IP per hour. Should be enough for regular usages.

Issue #379, opened 2024-05-29
Saving plans with result from precise location search yield unexpected results


timwangmusic:
When querying with precise location feature, and save the results to user profile leads to the title being the lat/lng of the query.

PR #380, opened 2024-05-30
Fix precise location display issue in user saved plans

timwangmusic:
## Description
The goal of this PR is to fix issue #379 . The cause is we relies on URL parameter when posting user plan views. And when using precise location, this parameter appears in the form of latitude and longitude. And this in turn got saved to user plans as travel destinations.

## Solution
* Use travel destination value from API instead of URL parameter.
* Also disable updating user favorites when using precise location feature.

## Testing
- [ ] Added new unit tests

## Checks

### DIFF
--- a/assets/js/search_results.js
+++ b/assets/js/search_results.js
@@ -154,10 +154,10 @@ rollUpButton.addEventListener("click", () => {
 
 $(document).ready(async function () {
   {
-    const plans = await getPlans();
-    for (let idx = 0; idx < plans.length; idx++) {
+    const data = await getPlans();
+    for (let idx = 0; idx < data.travel_plans.length; idx++) {
       let btn = document.getElementById("save-" + idx);
-      if (btn != null && plans[idx].saved) {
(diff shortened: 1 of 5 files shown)


assets/js/search_results.js
```diff
@@ -50,24 +50,25 @@ async function postPlanForUser() {
...
 
   $(this).attr("disabled", "true");
   $(this).parent().attr("title", "saved!");
 }
 
-function planToView(plan, planIndex) {
+function planToView(plan, planIndex, destination) {
   const url = new URL(document.URL);
-  const location = normalizeLocation(url.searchParams.get("location"));
+  const location = normalizeLocation(destination);
```
sananand007, line 71: nice.

sananand007: Do we need `url` anymore ?. I see we get the `date` ?

sananand007 (approved):
lg

assets/js/search_results.js
timwangmusic: that's correct.

PR #385, opened 2024-06-23
Fix: weekday value in planning requests

timwangmusic:
## Description
- Weekday transformation function from date in request is wrong. It doesn't compensate for the fact that the `time.Weekday` type starts with Sunday while the internal Weekday starts on Monday.

## Solution
* Compensate for the difference between `time.Weekday` type and the internal `Weekday` type.

## Testing
- The Louve Museum closes on Tuesday, currently the planner recommends the museum on Tuesday, which is obviously wrong.

## Checks


sananand007 (approved):
lg

PR #389, opened 2024-08-04
Reduce plans API calls in the search result page

timwangmusic:
## Description
As the functionality for the plan result page increases, we make increasing number of calls to the planning API. Currently there are four types of functionalities make calls to the API endpoint:
* when the page is loaded
* when a user saves a plan
* when a user requests a plan summary
* when a user posts the feedback for a plan

## Solution
* Store the plans data in a global variable so that all the functionalities can access them once the page is loaded.

## Testing
- [ ] Added new unit tests

## Checks


sananand007 (approved):
lg

PR #391, opened 2024-09-15
refactor: use Redis transactions

timwangmusic:
## Description
The former user favorite design was poor, due to the lack of consideration of concurrency issues. Problem arises when user access APIs from different devices which lead to inaccurate API request counts for a particular location.

Also in this change, we improve the location input handling. Now regardless of whether user enters "San Diego, USA" or "san diego, CA, United States", all these inputs will be mapped to "San Diego, CA, UNITED STATES" as a standard format.

## Solution
* Store user search history in a dedicated key in Redis.
* Use Redis transaction to watch the key while updating user search history.
* Correct location input with geo query and reverse query to get a standardized location info and then format the location as a string.

## Testing
- [x] Added new unit tests

## Checks


iowrappers/users.go
```diff
@@ -79,25 +79,90 @@ func (r *RedisClient) UpdateSearchHistory(ctx context.Context, location string,
...
-	userView.Favorites.SearchHistory[location] = data
+	return nil
+}
 
-	Logger.Debugf("updating user favorite for user %s: %+v", userView.ID, data)
-	return r.UpdateUser(ctx, userView)
+type Favorites struct {
+	MostFrequentSearch string `json:"most_frequent_search"`
+}
+
```
sananand007, line 125: lg, what are the two new functions doing now ? `UserFavorites` and `findTopLocation` ?

sananand007 (approved):
lgtm .. nice work .

iowrappers/users.go
timwangmusic: User favorite finds the most searched location of a user. From users perspective they will not see difference in the profile page.

PR #395, opened 2024-12-30
fix: image location for location input with two parts

timwangmusic:
## Description
When users use locations with only city followed by country (e.g. Paris, France), the image is stored under `/` with the current implementation. In this fix, the images will be stored in `country/cities/city_name` directory in the S3 bucket.

## Solution
* Add a simple validation in the image fetcher.
* Add a fixed filler for location inputs without `admin area one`.

## Testing
- [ ] Added new unit tests

## Checks


sananand007 (approved):
lg

PR #401, opened 2025-05-04
feat: add location image download button

timwangmusic:
## Description
Add a download button for users to download the location image.

## Solution
* Since we have S3 signed URL, we just need to add a new download button and assign the same URL to its `download` property. 
* Use material icons.
* Also trying to fix image display issue on smaller devices.

## Testing
~~[ ] Added new unit tests~~

## Checks


sananand007 (approved):
lg

PR #427, opened 2026-03-21
Improve plan generation with per-slot truncation and spatial clustering

tim-eternos:
## Summary
- **Per-slot candidate truncation**: After sorting places by score, each slot is truncated
  to the top 30 candidates (`MaxPlacesPerSlot`), significantly reducing the combinatorial
  search space without sacrificing plan quality since places are pre-sorted by `PlaceScore`.
- **Spatial clustering**: Places across all slots are grouped into geographic clusters
  (using `searchRadius / 3` as cluster radius). The solver runs independently on each
  spatial group and merges results, naturally producing geographically coherent plans
  with less travel time.
- Includes a fallback to the original non-clustered approach when spatial clustering
  yields no valid groups (e.g., when a cluster has an empty slot).

## Test plan
- [x] New unit tests for `groupPlacesBySpatialClusters` covering single cluster,
      multiple clusters, empty-slot filtering, and empty input
- [x] Existing `TestSolutionCandidateSelection` still passes
- [x] All tests in `planner/`, `test/`, and `matching/` packages pass
- [ ] Manual testing with real search queries to verify plan quality and geographic coherence


timwangmusic (approved):
lgtm

PR #437, opened 2026-07-21
Add brand keyword search and POST /v1/nearby-places endpoint

tim-eternos:
Nearby search was category-only (Visit/Eatery over 6 Google place types),
so brand-level queries like "Dunkin' near a coordinate" were not expressible.
This adds:

- Keyword (brand) support in PlaceSearchRequest: passes Keyword to Google
  Maps NearbySearch with the place type unset, and caches results under
  brand-scoped Redis geo keys (placeIDs:brand:{brandKey}) so cached brand
  searches stay separate from category buckets
- StrictNameMatch filtering: drops keyword results whose names don't match
  the brand (Google matches keywords against reviews and other content),
  applied before results reach the brand cache
- Brand-aware last-search-time freshness tracking per city
- POST /v1/nearby-places: accepts {brands[], location, radius, limit},
  reverse-geocodes once, fans out per-brand searches concurrently, and
  returns operational places per brand sorted by distance
- processLocation short-circuit when a location already has precise
  coordinates and city-level info, so fan-out callers geocode once

## Description
Clearly describe changes in your PR, new feature or bug fixes.

## Solution
* Describe your code changes in detail for reviewers as well as documentation purposes. 
* For bug fixes, explain the solutions and how they fix the issues.
* For new features, describe the high-level ideas.

## Testing
- [x] Integration testing on Heroku staging
- [x] Added new unit tests

## Checks
- [x] Have you removed commented code?
- [x] Have you used gofmt to format your code?

PR #438, opened 2026-07-22
Require authentication on POST /v1/nearby-places

tim-eternos:
## Why

`POST /v1/nearby-places` (#437) shipped unauthenticated — the `v1` group has no auth middleware and the handler did no check. Each request triggers a reverse geocode plus up to 25 concurrent Google Maps searches on a cold cache, so an open endpoint is a cost/abuse hole for a machine-facing API.

## What

- `getNearbyPlaces` now calls `UserAuthentication(ctx, user.LevelRegular)` before doing any work, following the repo's per-handler auth convention. PAT `Authorization: Bearer` header is checked first, JWT cookie as fallback — so browser sessions keep working.
- Auth failure returns **401** with the standard `{"error": ...}` object shape (consistent with the endpoint's other error responses, so API clients can always read the `error` field).
- New `planner/nearby_places_auth_test.go`: no credentials → 401; invalid token → 401; valid PAT (user + token minted against miniredis) → passes auth and reaches request validation (400 on zero location), with no Google Maps dependency.

## Client provisioning note

Callers now need a PAT: `POST /v1/create-token` with e.g. `{"name": "offerbee-backend", "expiration_duration": "8760h"}`. Note `time.ParseDuration` does not accept a `d` (days) unit despite the field comment's examples, and the default when omitted is only 5 minutes — use hours for long-lived service tokens, and calendar the expiry (PATs always expire).

### DIFF
--- a/planner/nearby_places_auth_test.go
+++ b/planner/nearby_places_auth_test.go
@@ -1,0 +1,71 @@
+package planner
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/weihesdlegend/Vacation-planner/test/redis_client_mocks"
+	"github.com/weihesdlegend/Vacation-planner/user"
+)
+
+// The /v1/nearby-places handler must reject unauthenticated requests before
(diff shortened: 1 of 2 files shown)


timwangmusic (approved):
lgtm

PR #442, opened 2026-07-28
Classify category-endpoint places by primary Google type

tim-eternos:
## Problem

`POST /v1/nearby-places-by-category` returns businesses tagged with the **searched** place type, not their real one. Google returns a place for a `type=restaurant` search if it merely has `restaurant` as a *secondary* type, so a live audit at Fremont Civic Center surfaced:

| Place | Returned type | Actually is |
|---|---|---|
| Seafood City Supermarket | `restaurant` | grocery |
| Cinemark Century Union Landing | `restaurant` | cinema |
| Bay View Golf Club / Golfland | `restaurant` | recreation |
| Loop Neighborhood Market | `cafe` | convenience/grocery |
| Foodmaxx | `supermarket` | not food |
| Ace Hardware | `store` | not food |

Downstream (OfferBee best-card), a supermarket tagged `restaurant` earns the user's **dining** multiplier — wrong card, wrong rate — and non-food results consume the app's per-category slots. Root cause: `parsePlacesSearchResponse` assigns the searched type, and `POI.Place` discarded Google's real `Types` array.

## Fix

Classify each place by its **primary** function and let it appear only in the category that function belongs to.

- **`POI.Place`** gains `Types []string`; `parsePlacesSearchResponse` populates it from `maps.PlacesSearchResult.Types` (additive; trip planning ignores it).
- **`POI.PrimaryLocationType(types)`** — first non-umbrella type (`food`, `point_of_interest`, `establishment`, …); Google lists the most specific type first.
- **`POI.ReclassifyForCategory(place, cat)`** — keep + re-tag when the primary type is one of the category's search types; drop when the primary function is elsewhere (it surfaces under its own category's search instead); keep unchanged when `Types` is empty (older cached records → no coverage regression).
- **`getNearbyPlacesByCategory`** maps results through `ReclassifyForCategory` before the day-closed filter + nearest-N cap.

Out-of-scope primary types (movie_theater, golf, convenience_store, hardware_store) have no matching category and are omitted for now. No consumer/client change needed — the corrected `LocationType` flows through unchanged.

## Result

- Eatery returns only real cafes/restaurants; a supermarket the food search matched is dropped from Food and appears under Shopping tagged `supermarket` (→ grocery reward).
- Robust to existing cache pollution: reclassification happens at response time from `Types`, so mis-bucketed places are corrected on read.

## Testing

`go build ./...`; `go test ./POI/... ./iowrappers/... ./planner/... ./matching/... ./test/...` pass, incl. new `TestPrimaryLocationType` and `TestReclassifyForCategory` (supermarket-dropped-from-Eatery, kept-under-Shopping, cafe→restaurant re-tag, cinema dropped, empty-Types kept).

Note: existing warm caches carry pre-fix records; the filter corrects responses regardless, and `redis-cli FLUSHDB` fully repopulates with accurate types.

### DIFF
--- a/POI/categories.go
+++ b/POI/categories.go
@@ -109,6 +109,57 @@ func ParsePlaceCategory(s string) (PlaceCategory, bool) {
 	}
 }
 
+// umbrellaLocationTypes are Google's generic feature types that describe almost
+// every place and say nothing about its primary function. They are skipped when
+// picking a place's primary type.
+// Note: "store" is intentionally NOT here — it is a meaningful Shopping type.
+var umbrellaLocationTypes = map[LocationType]bool{
+	LocationType("food"):              true,
+	LocationType("point_of_interest"): true,
+	LocationType("establishment"):     true,
+	LocationType("premise"):           true,
+	LocationType("geocode"):           true,
+	LocationType("political"):         true,
+}
+
+// PrimaryLocationType returns a place's primary Google feature type: the first
+// entry in its Types list that isn't a generic umbrella (food, point_of_interest,
+// …). Google lists the most specific type first, so this is the place's real
+// function (e.g. "supermarket" for a store the restaurant search also matched).
+// Returns "" when there is no meaningful type (empty/unknown Types).
+func PrimaryLocationType(types []string) LocationType {
+	for _, t := range types {
+		lt := LocationType(t)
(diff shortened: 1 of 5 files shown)


timwangmusic (approved):
lgtm

PR #443, opened 2026-07-28
Parallelize per-place-type Nearby Searches (cold-latency)

tim-eternos:
## Problem

Cold-cache latency for `/v1/nearby-places-by-category` is dominated by `extensiveNearbySearch` running each of a category's place types' Google Nearby Search calls **sequentially**. Shopping expands to 5 types, Wellness to 3, so a cold category is 3–5 serial Google round-trips. Measured: a 4-category cold request at a fresh coordinate took **~2.3 s** (warm is ~0.1 s).

## Fix

Split each pass into two phases:

- **Phase A** fetches every eligible place type's Nearby Search **concurrently**, bounded by the existing shared API semaphore (`ctx`-aware acquire, so a request timeout can't leak a goroutine). Each goroutine writes only its own result slot — no shared state is touched.
- **Phase B** processes the fetched responses **sequentially, in `placeTypes` order** — unchanged from before. Cross-type dedup (`placeMap`), the shared Place Details budget (`selectPlacesForDetails`), append order, and page-token/count updates are all identical to the sequential implementation.

**Output is byte-for-byte identical** (same set, same order). `extensiveNearbySearch` is shared with trip planning, which is therefore unaffected — only the wall-clock of the Google fan-out changes: a category's N searches now overlap into ~one round instead of N.

## Testing

`go build ./...`; `go test -race ./iowrappers/... ./planner/... ./matching/...` (no data races); full `go test ./...` pass.

Live latency confirmation (rebuild + Maps key) pending, but the expected effect is cold ≈ 2.3 s → ~0.7–1 s (bounded by the 5-slot API semaphore and Google QPS).

Independent of #442 (place-type classification) — different functions, no conflict.

### DIFF
--- a/iowrappers/nearby_search.go
+++ b/iowrappers/nearby_search.go
@@ -146,25 +146,59 @@ func (c *MapsClient) extensiveNearbySearch(ctx context.Context, maxRequestTimes 
 	urlMap := make(map[string]string)       // map place ID to url
 	summaryMap := make(map[string]string)   // map place ID to summary
 
-	var err error
 	var mapsFailuresCount uint = 0
 	detailsBudget := request.DetailsLimit // remaining Place Details calls across all pages; only enforced when DetailsLimit > 0
+
+	// One place type's Nearby Search HTTP result (Phase A). Holds no shared state,
+	// so goroutines can fill it by index without synchronization.
+	type fetchResult struct {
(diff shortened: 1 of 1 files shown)


timwangmusic (approved):
LGTM

PR #450, opened 2026-08-01
Cut Google Maps spend: reverse-geocode cache, Details mask trim, confirm skip, closure persistence

timwangmusic:
The no-wire-change batch from the 2026-08-01 geo cost/freshness audit. Four cost cuts plus one zero-cost freshness fix:

## 1. Reverse-geocode cache (the big one)
Every nearby scan buys a reverse-geocode via `processLocation` — and on a warm place cache it was the **only** Google call the scan made, so it was ~100% of warm-scan spend. Results are now cached per ~8 km search cell (`geocode:reverse:<cell>`, 30-day expiry per Geocoding caching terms). The `RedisClient.ReverseGeocode` stub is now a real implementation; `processLocation`, `PoiSearcher.ReverseGeocode`, and the `/v1/reverse-geocoding` handler all route through it.

## 2. Details field-mask trim
`detailed_search_fields` drops `name` and `user_ratings_total`: no Details consumer reads either (both arrive free with every Nearby/Text Search result), and `user_ratings_total` alone pulls every Details call into the Atmosphere billing tier. The `AddUserRatingsTotal` migration passes its own single-field list and is unaffected. `editorial_summary` stays (trip planner reads it) — note that means the Atmosphere tier still applies until that's revisited. Pinned by `TestDetailedSearchFieldsMask`.

## 3. Confirm skips redundant Details
`/v1/place-search/confirm` bought a full max-tier Place Details call even when the cached record's details were current — data `restoreCachedDetails` restores anyway. It now applies the same `placeDetailsAreCurrent` rule the nearby path already trusts. This also makes Offerbee's wallet-change re-confirm of manual rows free in the common case.

## 4. Photos-only mask in stale-photo recovery
`photos_client.go` requested the entire Details mask to recover one photo reference; now `["photos"]`.

## 5. Closures are persisted (freshness fix, zero cost)
Cold searches filtered non-`OPERATIONAL` results **before** the cache write, discarding the closure signal entirely — a permanently closed place kept its `OPERATIONAL` record and cache membership forever. The write now includes closures (new `persistAndFilterSearchResults` owns write + response filter); the response-side behavior is unchanged, and the existing read-side Operational filter retires closed places from cache serves.

## Testing
- TDD throughout: new suites `reverse_geocode_cache_test.go` (round-trip, cell scoping, expiry, cache-first searcher + processLocation), `add_searched_place_currency_test.go` (skip-when-current, buy-when-stale, buy-when-uncached), `closure_persistence_test.go` (persist + response exclusion + read-path retirement), `config_yaml_test.go` (mask pin).
- `go vet ./...` clean; full `go test ./...` green (7 packages).
- The photos one-liner has no unit test — that call site has no stub seam; covered by build + the narrowed literal.

Deliberately not in this PR: `hoursKnown` wire flag (API change), text-search caching + timeout salvage (larger), `MapsLastSearchTime` dead-field purge (Redis ops).

https://claude.ai/code/session_01DjdGnZM6cUBeg6jWwoQfU3

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -1,12 +1,14 @@
 server:
   google_maps:
+    # Every field is billed per Details call, tiered by field group — see
+    # TestDetailedSearchFieldsMask for why name and user_ratings_total are
+    # deliberately absent (nothing reads them from a Details response; both
+    # arrive free with every Nearby/Text Search result).
     detailed_search_fields:
-      - name
       - opening_hours
       - formatted_address
       - adr_address
       - url
-      - user_ratings_total
       - editorial_summary
       - photos
   plan_solver:
--- a/config_yaml_test.go
+++ b/config_yaml_test.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"os"
+	"testing"
+
+	"gopkg.in/yaml.v3"
+)
+
+// Pins the Place Details field mask in config/config.yml. Every field here is billed per Details
+// call, tiered by field group (Basic / Contact / Atmosphere on the legacy API), so an unused
+// field is pure spend:
+//   - name and user_ratings_total must stay OUT: neither is read from a Details response on any
+//     path (both already arrive free with every Nearby/Text Search result), and
+//     user_ratings_total alone pulls the call into the Atmosphere tier. The AddUserRatingsTotal
+//     admin migration passes its own single-field list and is unaffected.
+//   - the remaining fields are load-bearing: opening_hours (open-now filtering),
+//     formatted_address/adr_address (display + address parsing), url (also the Details-freshness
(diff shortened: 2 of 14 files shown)


timwangmusic:
Added a fail-safe flag on top of the closure-persistence change: `PlaceSearchRequest.BusinessStatus` is replaced by `IncludeClosedPlaces` (zero value = non-Operational places always filtered, from cache reads and cold-search responses alike). Previously a caller had to remember `BusinessStatus: POI.Operational` to get filtering — with closures now actually present in the cache, a forgotten field would have served closed places to planners. All three production call sites already requested filtering, so behavior is unchanged; only the failure mode is gone.

https://claude.ai/code/session_01DjdGnZM6cUBeg6jWwoQfU3

tim-eternos (approved):
lgtm

PR #452, opened 2026-08-12
Derive LocationType from the primary Google type for keyword searches

timwangmusic:
## Problem

Keyword (brand) searches (`/v1/nearby-places`) query Google with `LocationTypeAny`, and `parsePlacesSearchResponse` stamps every result with the *searched* type — so keyword results are cached with `LocationType: ""`. The write path (`SetPlacesAddGeoLocationsForBrand`) is a blind upsert into the shared `placeDetails:{id}` records that category reads hydrate from, and `restoreCachedDetails` restores URL/Summary/Address/Hours but not LocationType. Net effect: every brand search wipes the type off previously typed records.

Observed downstream (OfferBee staging): a McDonald's row cached by an Eatery category search was served with `LocationType: ""` after brand searches rewrote its record, so the best-card ranking scored every card at its base rate — Chase Sapphire Reserve showed 1X at McDonald's instead of 3X dining.

## Fix (both changes gap-filling only)

- **`parsePlacesSearchResponse`**: when the searched type is `LocationTypeAny`, derive `LocationType` from `POI.PrimaryLocationType(place.Types)` — the same machinery `ReclassifyForCategory` already uses to re-tag places on category reads (#442). Google lists the most specific type first and umbrella types (`food`, `point_of_interest`, …) are skipped, so a McDonald's (`["meal_takeaway", "food", ...]`) comes out `meal_takeaway`. Typed searches keep their searched-type stamp — re-tagging those stays `ReclassifyForCategory`'s job.
- **`restoreCachedDetails`**: restore the stored `LocationType` when the rebuilt record has none — same "only ever fills a gap, never overwrites" rule the function already documents for the Details-sourced fields. Covers keyword results whose `Types` carry no meaningful primary.

## Safety

- Category geo buckets can't be poisoned: the brand write path never touches them, and `SetPlacesAddGeoLocations` already refuses unmapped types ("Refuse to guess a bucket").
- Empty/umbrella-only `Types` → `PrimaryLocationType` returns `""` → identical to today.
- Existing blank-typed records stay blank until a fresh search rewrites them; this stops new wipes at the source. (OfferBee's section-fallback covers the residue meanwhile.)

## Testing

- New: `TestParsePlacesSearchResponseKeywordDerivesLocationType` (keyword derives from Types; umbrella-only stays blank; typed search stamp unchanged), `TestRestoreCachedDetailsPreservesLocationType` (untyped rebuild recovers stored type; fresh type never overwritten). Both watched fail before the fix.
- `go build ./... && go vet ./iowrappers/ && go test ./...` all green.

https://claude.ai/code/session_011BmBJ2x6b7V36vywSMrYDy

### DIFF
--- a/iowrappers/nearby_search.go
+++ b/iowrappers/nearby_search.go
@@ -364,6 +364,9 @@ func restoreCachedDetails(places []POI.Place, cached map[string]POI.Place) {
 		if places[i].Address == (POI.Address{}) {
 			places[i].Address = stored.Address
 		}
+		if places[i].LocationType == POI.LocationTypeAny {
+			places[i].LocationType = stored.LocationType
+		}
 		if !places[i].HasRealOpeningHours() && stored.HasRealOpeningHours() {
 			places[i].Hours = stored.Hours
 		}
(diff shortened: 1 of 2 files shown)


Ronnie434 (approved):

tim-eternos (approved):
lgtm

PR #454, opened 2026-08-22
config: drop editorial_summary and photos from the Details field mask

timwangmusic:
## Why

The production GCP instance already runs with these two fields disabled as a local config override — but master's committed config still requests them, and `TestDetailedSearchFieldsMask` pinned them as *required*. Any redeploy from a fresh checkout would silently resurrect the spend, and nothing but the Google invoice would say so.

`editorial_summary` alone pulls every Place Details call into the Atmosphere billing tier; `photos` feeds the photo client's separately-billed image fetches. The main consumer driving Details volume (OfferBee's nearby/category/search paths) reads neither.

## What

- `config/config.yml`: `detailed_search_fields` trimmed to the load-bearing four — `opening_hours`, `formatted_address`, `adr_address`, `url`.
- `config_yaml_test.go`: `editorial_summary` and `photos` move from the required list to the banned list, with the rationale in the pin's doc comment — re-adding either is a deliberate spend decision, not a bug fix.

## Safety

Every reader of both fields is nil-guarded (`nearby_search.go:410/490`, `text_search.go:347/354`, `photos_client.go:151`), so places simply carry no summary/photo — the behavior production has already had since the instance-side disable, now made durable.

## Testing

Test-first: the updated pin failed against the old config (both fields flagged), passed after the trim. Full suite green: root, `iowrappers`, `planner`, `test/...`.

https://claude.ai/code/session_01YF2pmkEjcz5Uqh2sRgzypQ


timwangmusic:
Wrong repo — OfferBee's geo work lives on offerbee-ai/Vacation-Planner, where this exact trim already shipped 2026-08-13 (PR #10). Closing as redundant.

